PR #2, opened 2019-07-25
Set git quotepath off

mehdisfdc:
This commit disables default git behaviour where non-ASCII file names are printed in quoted octal notation. This default behaviour would cause the script to fail if file names contain special characters.


scolladon:
@mehdisfdc please rebase this branch on top of master in order to resolve conflict.
I think unit test will pass because there is already a mock for ```child_process.spawnSync``` used in the related test file

PR #34, opened 2020-06-27
Fix non existing source

### LINKED ISSUE #33: ENOENT no such file or directory *.undefined
Hello @scolladon 

A new one, unable to find a file that is not in my repo.

I have checked, the whole object ContractedPrice deployment is ok  with sfdx force:source:deploy so no error in it I guess.

Reproduction step : 
```
$  sgd -d -t HEAD -f origin/master -r . -o output
Error terminating the process during sfdx-git-delta : Error: ENOENT: no such file or directory, stat 'force-app\main\default\objects\DWIN2_ContractedPrice__c\DWIN2_ContractedPrice__c.undefined'

$ git diff HEAD^..origin/master --stat | grep Contract
 .../DWIN2_ContractedPrice__c.object-meta.xml       |    2 +-
 .../fields/DWIN2_Contract__c.field-meta.xml        |    2 +-
```

scolladon:
Closes #33 
# Breaking changes
- return status code 1 when non recuperable errror are catched, else return 0 (even if warnings are catched)
- return a json object in every case :
```javascript
{
  "error": null, // <error object>, if any
  "output": "output", // <path>, output folder
  "success": true, // <boolean>, success or failure
  "warnings": null // <array of error object>, any recuperable error are here
}
```


bin/cli
```diff
@@ -27,11 +27,21 @@ program
...
+  const jobResult = orchestrator(program)
+  if (jobResult.warnings && jobResult.warnings.length) {
+    output.warnings = jobResult.warnings
+  }
   process.exitCode = 0
 } catch (err) {
-  console.error(`Error terminating the process during sfdx-git-delta : ${err}`)
+  output.success = false
+  output.error = err
   process.exitCode = 1
```
alex9138, line 45: 👍

PR #36, opened 2020-07-01
Add the corresponding line to the warning message

scolladon:
Improve message content when displaying warning to help understand the root cause of the warning


scolladon:
closes #37

PR #52, opened 2020-09-07
Bugfix/fix incorrect metadata handling

jjulicher:

---

Fixes to handle customweblink and where the name of custom metadata (or any metadata type) contains it's own suffix. e.g. if a custom metadata name starts with .md it would cause issues (see example in test).

- [ ] Added for new features.
- [x] Fixed for any bug fixes.

## Explain your changes

---

Bugfixes to handle custompageweblink & custommetadata names starting with 'md'

No.
---

_(Write your answer here.)_

---

_(Write your answer here.)_

## Any other comments?

---

_(Write your answer here.)_

On my local and CI/CD server. 
---

**Operating System:** …
Windows 10. 
**NPM version:** …
6.14.8
**Node version:** …
14.9
**sgd version:** …
3.31
**git version:** …
2.17.1

### DIFF
--- a/__tests__/unit/lib/service/standardHandler.test.js
+++ b/__tests__/unit/lib/service/standardHandler.test.js
@@ -20,6 +20,16 @@ const testContext = {
       'force-app/main/default/quickActions/NewGlobal.quickAction-meta.xml',
       new Set(['NewGlobal']),
     ],
+    [
+      'customMetadata',
+      'force-app/main/default/customMetadata/GraphicsPackImages.md_png.md-meta.xml',
+      new Set(['GraphicsPackImages.md_png']) // previously it would come out as GraphicsPackImages_png
+    ],
+    [
+      'weblinks',
+      'force-app/main/default/weblinks/ClientStore.custompageweblink-meta.xml',
+      new Set(['ClientStore']) // previously it would come out as GraphicsPackImages_png
+    ],
   ],
   work: {
     config: { output: '', repo: '', generateDelta: true },
--- a/__tests__/unit/lib/service/standardHandler.test.js
+++ b/__tests__/unit/lib/service/standardHandler.test.js
@@ -23,12 +23,12 @@ const testContext = {
     [
       'customMetadata',
       'force-app/main/default/customMetadata/GraphicsPackImages.md_png.md-meta.xml',
-      new Set(['GraphicsPackImages.md_png']) // previously it would come out as GraphicsPackImages_png
+      new Set(['GraphicsPackImages.md_png']),
     ],
     [
       'weblinks',
-      'force-app/main/default/weblinks/ClientStore.custompageweblink-meta.xml',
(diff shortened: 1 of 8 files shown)


scolladon:
Hi @jjulicher !

Thanks for your work !

I open the discussion here :) 

I'm not sure for the custompageweblink suffix.
It seems in the [documentation](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/custompageweblink.htm) that webLink is used anyway.
Plus, webLink should still be working and it seems to me that with this PR only custompageweblink will be working now.
What do you think @mehdisfdc ?

Am I missing something ?

scolladon (commented):
I am ok for all the files I have looked.

For custompageweblink I think we should discuss more

scolladon (commented):
I am ok for all the files I have looked.

For custompageweblink I think we should discuss more

scolladon:
After reading the documentation here is what I understand.

CustomPageWebLink are metadata representing Custom Button created for classic Home Page Component. They are stored in a webLinks folder at the package root directory.

WebLink are metadata related to CustomObject. In the metadata format it is contained in the CustomObject file. In the source format it is contained in a sub folder of the CustomObject.

This is a little bit annoying, as they both have the same parent folder name (but not at the same level in the package directory), because the handler selection is done based on the folder name... So with the current implementation it is not possible without changing the code for this particular to make it works for both.

I wonder if we should implement support for CustomPageWebLink part of this PR as this is a metadata related to classic and not used in Lightning anymore because home page are flexipage now.

Do you guys @mehdisfdc & @jjulicher understand it the same way ? 
Am I missing something ?

jjulicher:
The basics of it were:

The CustomWebLinks members were being prefixed with 'default'. 

This update removed the prefix and allowed a deploy to happen - otherwise it would error. 

If you want to implement it another way sure that's cool but as it stood it would fail my deployments.

scolladon:
@jjulicher Can you explain what you have created to have custompageweblink in your org and then how you retrieve them in your repository please ?

jjulicher:
> @jjulicher Can you explain what you have created to have custompageweblink in your org and then how you retrieve them in your repository please ?

I used `sfdx force:source:retrieve`

and 	package.xml had: 

```
	<types>
		<members>*</members>
		<name>CustomPageWebLink</name>
	</types>
```

As for creating them in the org - not sure - I'm working with a legacy org at the moment and was just pulling down everything and trying to push everything into a sandbox in a CI pipeline.

scolladon:
Ok I get it thanks.

So here is what I suggest : 

1. Keep the regex changes on the files lib/service/inFolderHandler.js and lib/service/standardHandler.js (because they fix something)
2. Move the CustomPageWebLinks part of this PR in an issue of type enhancements. We will be able to discuss about it there and see if there is traction around this feature as it is related to classic component.

@jjulicher could you do that ?

scolladon (approved):

mehdisfdc (approved):

PR #53, opened 2020-09-07
Bugfix/report types incorrect type handler

jjulicher:

---

- [ ] Added for new features.
- [x] Fixed for any bug fixes.

## Explain your changes

---

Set reportTypes to use standard handler as when generating the package.xml a report type comes out as default.Report_Type_Name - which won't deploy.

Added .idea into .gitignore to ignore JetBrain IDE settings. 

---

no

---

_(Write your answer here.)_

## Any other comments?

---

_(Write your answer here.)_

Tested on local box using sfdx cli after package generation. 
---

**Operating System:** …
windows 10/ wsl ubuntu 20.04
**NPM version:** …
6.14
**Node version:** …
14.9
**sgd version:** …
3.31
**git version:** …
2.17.1

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,3 @@
 node_modules
 coverage
 output
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,0 @@
-node_modules
-coverage
-output
-.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 node_modules
 coverage
 output
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+node_modules
+coverage
+output
+.idea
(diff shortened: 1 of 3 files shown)


scolladon:
Hi @jjulicher !

Thanks for your work!

That's a very nice catch for the report types. I am going to make some tests locally soon.

Could remove the gitignore commit please ? development file are added locally, we just don't keep track of build artefact
In this case I suggest you to use the local ignore git feature [.git/info/exclude](https://git-scm.com/docs/gitignore)

scolladon (changes requested):
Thanks for this PR, nice catch
Please remove the change on the gitignore for the reasons stated in comment
Then I'll validate the PR

.gitignore
```diff
@@ -1,3 +0,0 @@
-node_modules
-coverage
-output
```
scolladon, line 3: Let those lines, they are to ignore in the development workflow of the project
Just don't add files related to IDE here, put them locally (we don't need to share them)

scolladon (changes requested):
Do not change .gitignore please (not modification neither deletion)
Let it as it is

scolladon (approved):

PR #65, opened 2020-09-30
Move to sfdx plugin 

scolladon:

---

- [ ] Added for new features.
- [x] Changed for changes in existing functionality.
- [ ] Fixed for any bug fixes.

## Explain your changes

---

Migrate from commander cli to sfdx plugin oclif structure
Migrate all the tool chain from yarn to npm to follow sfdx plugin coding style
The command line keeps the old cli (sgd) and also add the ability to include it as a sfdx-plugin

the new command is now `sfdx force:source:delta ...` with the same parameters

### Breaking change !

The behavior of the -i parameter is not the same due to the pedantic of the oclif framework

Value matrice :
if -i is not set then ignore is undefined
if -i is set with no value then error
if -i is set with a value then ignore equals this value

=> No more default value to .forceignore when setting -i without value.

No
- [x] Jest test to check the fix is applied are added.

execute `bin/run force:source:delta` locally to see how it goes
Follow the best practices to test it locally via [plugin linking](https://developer.salesforce.com/docs/atlas.en-us.sfdx_cli_plugins.meta/sfdx_cli_plugins/cli_plugins_debug.htm)

---

**Operating System:** Mac Os Mojave 10.14.6

**Yarn version:** 1.22.5 

**Node version:** v14.12.0

**sgd version:** 3.4.0

**git version:** 2.28.0


scolladon:
@mehdisfdc you can do your magic in the README :)

mehdisfdc (approved):
Can't wait to start using the Salesforce CLI plugin next week 😃

PR #139, opened 2021-04-29
docs: add mention to yq

### LINKED ISSUE #126: Make List of Apex Class Names from Delta Package.xml to Use In Delta Deployment with Test Run
---
It is fantastic, that the SGD generates a delta package.xml, which is used to deploy the delta. However, if the delta package.xml file contains apex classes (and apex test classes), there is no easy way to generate a list of apex classes to use for the RunSpecifiedTest part of a validation run

If the SGD plugin also could generate such a list, it would make it so easy to also run selected tests in a validation run, further enhancing the delta deployment experience. 

E.g. the command for running specified tests: 
`sfdx force:source:deploy --testlevel RunSpecifiedTests --runtests 'MyApexClass1, MyApexClass1_Test, ... '`

---
If the SGD command generates the following package.xml file:
```xml
    <types>
        <members>MyApexClass1</members>
        <members>MyApexClass1_Test</members>
        <members>...</members>
        <name>ApexClass</name>
    </types>
    <version>51.0</version>
</Package>
```
Then I want the SGD plugin to also generate a .txt file (e.g. ApexClassNames.txt) or return a string. The file/string should contain the names of apex classes in above package.xml file (both normal classes and test classes), separated by commas.
E.g. content is string with value `MyApexClass1, MyApexClass1_Test, ... `

If .txt file, it can be placed in a new folder, name is optional, to be decided.

---
The alternative is to generate a script, which gets the package.xml file from the SGD method, access the type with name ApexClass and retrieve all members from the section. Not the prettiest solution, and script will vary between each CI/CD developer, who has this need.

mehdisfdc:
## Explain your changes

Add a mention to the yq tool in the README, as a solution to transform the output from SGD.

closes #126

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,7 +262,7 @@ Depending on your testing strategy, [you may be interested in generating a a com
 
 To cover this requirement, you can use a tool such as [yq](https://github.com/kislyuk/yq) to parse the content of the package.xml file produced by SGD:
 
-`cat package/package.xml | xq . | jq '.Package.types | select(.name=="ApexClass") | .members | join(",")'`
+`xq . < package/package.xml | jq '.Package.types | if type=="array" then .[] else . end | select(.name=="ApexClass") | .members | join(",")'`
 
 ## Javascript Module
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # SFDX-Git-Delta ![Actions Status](https://github.com/scolladon/sfdx-git-delta/workflows/CI/badge.svg) [![Maintainability](https://api.codeclimate.com/v1/badges/95619399c7bb2cf60da4/maintainability)](https://codeclimate.com/github/scolladon/sfdx-git-delta/maintainability) [![Test Coverage](https://api.codeclimate.com/v1/badges/95619399c7bb2cf60da4/test_coverage)](https://codeclimate.com/github/scolladon/sfdx-git-delta/test_coverage) [![Known Vulnerabilities](https://snyk.io//test/github/scolladon/sfdx-git-delta/badge.svg?targetFile=package.json)](https://snyk.io//test/github/scolladon/sfdx-git-delta?targetFile=package.json)
 
-Generate the sfdx content in source format and destructive change from two git commits
+Generate the sfdx content in source format and destructive change from two git commits.
 
 ## What is SFDX-Git-Delta?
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -254,7 +254,15 @@ $ sfdx sgd:source:delta --from commit --ignore-destructive destructiveignore
 
 ```
 
-Note that in a situatrion where only the `--ignore [-i]` parameter is specified (and `--ignore-destructive [-D]` is not specified), then the plugin will ignore items matching `--ignore [-i]` parameter in all situations: Addition, Modification and Deletion.
+Note that in a situation where only the `--ignore [-i]` parameter is specified (and `--ignore-destructive [-D]` is not specified), then the plugin will ignore items matching `--ignore [-i]` parameter in all situations: Addition, Modification and Deletion.
```
scolladon, line 257: **praise**: nice catch for the "situation" spelling error, thanks

README.md
```diff
@@ -254,7 +254,15 @@ $ sfdx sgd:source:delta --from commit --ignore-destructive destructiveignore
...
+
+#### Generate a comma-separated list of the added and modified Apex classes:
+
+Depending on your testing strategy, [you may be interested in generating a a comma-separated list of the added and modified Apex classes](https://github.com/scolladon/sfdx-git-delta/issues/126) (to use in the `sfdx force:source:deploy --testlevel RunSpecifiedTests` command, for example).
+
+To cover this requirement, you can use a tool such as [yq](https://github.com/kislyuk/yq) to parse the content of the package.xml file produced by SGD:
+
+`cat package/package.xml | xq . | jq '.Package.types | select(.name=="ApexClass") | .members | join(",")'`
```
scolladon, line 265: **suggestion**(_non-blocking_): we could consider using "<" instead of cat

something like that (not tested, not sure it works this way, but the idea is here) : 
`xq . < package/package.xml | jq '.Package.types | select(.name=="ApexClass") | .members | join(",")'`

scolladon (approved):

README.md
mehdisfdc: I like your version! 
However (regardless of the use of < or cat) I've just realized that `.Package.types` won't work when there are several types in the package.xml (`.Package.types[]` would be need then). I'd have to come with a different syntax, please don't merge the PR for now.

scolladon (approved):

PR #185, opened 2021-08-19
feat: permissive git diff parameter

### LINKED ISSUE #184: [permissivediff] Allow to use permissive git diff to ignore diffs related to indentation

sfdx plugin [sfdx-hardis](https://www.npmjs.com/package/sfdx-hardis) uses sfdx-git-delta to:
- update local package.xml after a force:source:pull
- manage a file packageDeployOnChange.xml , to avoid deploying items whose state have not changed since last deploy

Manage a new argument `--permissivediff` to add the following parameters to **git diff** call 
```javascript
[
    '--ignore-all-space',
    '--ignore-blank-lines',
    '--ignore-cr-at-eol',
    '--word-diff-regex',
]
```

It allows to avoid detecting as a difference the indentation/spaces stuff

None

### Additional context

I almost completed the dev in my fork of sfdx-git-delta, a PR will come soon

nvuillam:
Manage a new argument --permissivediff to add the following parameters to git diff call

```javascript
[
    '--ignore-all-space',
    '--ignore-blank-lines',
    '--ignore-cr-at-eol',
    '--word-diff-regex',
]
```

It allows to avoid detecting as a difference the indentation/spaces stuff

Fixes #184

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,22 +93,22 @@ If you encounter this issue while having installed the correct version of node o
 ## How to use it?
 
 <!-- commands -->
-* [`sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-P] [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]`](#sfdx-sgdsourcedelta--f-string--t-string--r-filepath--i-filepath--d-filepath--s-filepath--p--o-filepath--a-number--d---json---loglevel-tracedebuginfowarnerrorfataltracedebuginfowarnerrorfatal)
+* [`sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-W] [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]`](#sfdx-sgdsourcedelta--f-string--t-string--r-filepath--i-filepath--d-filepath--s-filepath--w--o-filepath--a-number--d---json---loglevel-tracedebuginfowarnerrorfataltracedebuginfowarnerrorfatal)
 
-## `sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-P] [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]`
+## `sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-W] [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]`
 
 Generate the sfdx content in source format and destructive change from two git commits
 
 ```
 USAGE
-  $ sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-P] 
+  $ sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-W] 
   [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel 
   trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]
 
 OPTIONS
   -D, --ignore-destructive=ignore-destructive                                       ignore file to use
 
-  -P, --permissive-diff                                                             ignore git diff white char (space,
+  -W, --ignore-whitespace                                                           ignore git diff whitespace (space,
                                                                                     tab, eol) changes
 
   -a, --api-version=api-version                                                     [default: 52] salesforce API version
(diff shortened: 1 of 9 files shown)


scolladon:
Hi @nvuillam !

_I copy paste my comment from the related issue_

I have questions:

- Why this should not be the standard behaviour? => What will be the added value to that and why both use cases should exists?
- why using `--word-diff-regex` parameter? With which value?

Suggestion: the parameter name must follow the [cli parameter apache convention](https://commons.apache.org/proper/commons-cli/index.html). I suggest the following `--permissive-diff -P` if we keep this wording

Please add me to your fork so I can contribute and help you solve CI/CD issues you may encounter

nvuillam:
@scolladon I see you made the code more clean, nice ! :)

When do you think this could be released ? It would greatly help sfdx-hardis ^^

scolladon:
Hi @nvuillam !
 
> When do you think this could be released ? It would greatly help sfdx-hardis ^^
We are going to discuss about it today with @mehdisfdc and see how we can improve (if we can) and when to ship it
We'll keep you posted!

nvuillam:
That's great, thanks for the feedback :)

README.md
```diff
@@ -93,20 +93,24 @@ If you encounter this issue while having installed the correct version of node o
...
-  $ sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-o 
-  <filepath>] [-a <number>] [-d] [--json] [--loglevel 
+  $ sfdx sgd:source:delta -f <string> [-t <string>] [-r <filepath>] [-i <filepath>] [-D <filepath>] [-s <filepath>] [-P] 
+  [-o <filepath>] [-a <number>] [-d] [--json] [--loglevel 
   trace|debug|info|warn|error|fatal|TRACE|DEBUG|INFO|WARN|ERROR|FATAL]
 
 OPTIONS
   -D, --ignore-destructive=ignore-destructive                                       ignore file to use
+
+  -P, --permissive-diff                                                             ignore git diff white char (space,
```
mehdisfdc, line 111: @scolladon As discussed together, would you prefer to name the new command -`W, --ignore-whitespace` instead of `-P, --permissive-diff`? Both are fine with me.
@nvuillam Any opinion on this alternative name?

src/utils/gitConstants.js
```diff
@@ -3,5 +3,11 @@ module.exports.ADDITION = 'A'
 module.exports.DELETION = 'D'
 module.exports.GIT_DIFF_TYPE_REGEX = /^.\s+/u
 module.exports.MINUS = '-'
+module.exports.PERMISSIVE_DIFF_PARAMS = [
+  '--ignore-all-space',
+  '--ignore-blank-lines',
+  '--ignore-cr-at-eol',
+  '--word-diff-regex',
```
mehdisfdc, line 10: @nvuillam According to the [git documentation](https://git-scm.com/docs/diff-options#Documentation/diff-options.txt---word-diff-regexltregexgt), the `--word-diff-regex` parameter expects a `<regex>` value.
Otherwise it may apply a regex that is set locally in your gitattributes or git-config (extract from doc: "_The regex can also be set via a diff driver or configuration option, see gitattributes[5] or git-config[1]. Giving it explicitly overrides any diff driver or configuration setting._").

To make sure the behaviour of this command is consistent for all SGD users, what do you think about specifying explicitly the regex to use? In other words, I suggesting to modify line 10 to specify explicitly the `<regex>` in `--word-diff-regex=<regex>` (I guess you could look into you gitattributes and git-config to retrieve the value that you used during your tests).

@scolladon To go further, in addition to specifying explicitly the default regex to apply, we could even offer the ability for the users to override it with their own regex (but that would a "nice-to-have" feature, while ensure a consistent behaviour in the command in more important IMHO).

scolladon: @mehdisfdc I love the id to let the user override the default regex, I will implement it with the new parameter name (once everyone is ok) and with the default regex value

README.md
nvuillam: @mehdisfdc --ignore-whitespace-diff maybe ? :)
Anyway, while the capability is here, I don't matter the argument name :)

src/utils/gitConstants.js
nvuillam: More flexibility for the user, I fully agree, great suggestion :)

README.md
scolladon: I like `-W, --ignore-whitespace` 👍

src/utils/gitConstants.js
scolladon: I bounce on @mehdisfdc's message, @nvuillam could you give us the git config taken by `--word-diff-regex`?

nvuillam: I'm sorry, I just used the parameter and I think I viewed a difference compared to without, but i've not sent any value to this parameter

scolladon: Could you retry on your side to make sure it does something ?
And if it does something on your side, as you don't use any regex with it, could you send us your gitattributes and git config please ?

nvuillam: I don't have any .gitattributes defined

About git config, except user.name & user.email, I use 
- core.longpaths=true
- core.quotepath=false

I'll try to make new tests, but unfortunately my work week is already busy, it may wait until next week :/

nvuillam: But why not publishing as it is now, and send --word-diff-regex to git if --word-diff-regex is sent to sfdx-git-delta, and append --word-diff-regex value only if it is sent ? :)

scolladon: I prefer not to do that because if no parameter are given to `--word-diff-regex` then it uses the local configuration in `.gitattributes` and `.git/config`, which can be anything and could lead to unexpected behaviour

scolladon: So, I poced around and play a little bit with it.
It is also true for me that `--word-diff-regex` is required to have the right behavior.
I dig a little bit more in the documentation of this parameter in git and I found that there are [default regex](https://stackoverflow.com/questions/30428377/what-are-git-diff-word-diff-default-regexps) used (which are happened to the override foundable in `.gitattributes` and `.git/config`)

I wonder if we will have issues about this and how it will be to debug 😅 but I don't think it will be used by most of the users.
With that knowledge, it looks good to me 😄

mehdisfdc (approved):

scolladon (approved):

PR #216, opened 2021-12-03
doc: fix typo

mehdisfdc:


scolladon:
This has been fused directly into last main commit to keep a lean history
👍

PR #247, opened 2022-02-22
feat: output folder will be created if it doesn't exist

denisab85:
## What does this pull request contains

---

- [x] Added for new features.
- [ ] Fixed for any bug fixes.

## Explain your changes

---

---

closes #

---

## Any other comments?

---

The old test for the requirement of the --output folder to exist was obsoleted as unnecessary.
---

**Operating System:** …

**yarn version:** …

**node version:** …

**git version:** …

**sfdx version:** …

**sgd plugin version:** …

### DIFF
--- a/__tests__/unit/lib/utils/cliHelper.test.js
+++ b/__tests__/unit/lib/utils/cliHelper.test.js
@@ -47,16 +47,6 @@ describe(`test if the appli`, () => {
     }).toThrow()
   })
 
-  test('throw errors when output folder does not exist', () => {
-    cliHelper = new CLIHelper({ ...testConfig, to: undefined })
-    expect(() => {
-      cliHelper.validateConfig({
-        ...testConfig,
-        output: 'not/exist/folder',
-      })
-    }).toThrow()
-  })
-
   test('throw errors when output is not a folder', () => {
     cliHelper = new CLIHelper({ ...testConfig, output: 'file' })
     expect(() => {
--- a/src/utils/cliHelper.js
+++ b/src/utils/cliHelper.js
@@ -23,9 +23,13 @@ class CLIHelper {
     if (isNaN(this.config.apiVersion)) {
       errors.push(`api-version ${this.config.apiVersion} is not a number`)
     }
-    ;[this.config.output, this.config.source]
-      .filter(dir => !dirExist(dir))
-      .forEach(dir => errors.push(`${dir} folder does not exist`))
+    if (!dirExist(this.config.source)) {
+      fs.mkdir(this.config.source, { recursive: true }, err => {
+        if (err) {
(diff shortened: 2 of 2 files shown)


scolladon:
Hi @denisab85 !

Thanks for this contribution, that is a good idea.

We thought about it in the past with @mehdisfdc.
And we decided to not create the output folder if it does not exist because we don't want the tool to be used in order to create folder where it should not.
The tool can be installed using super user privileged and then be used by regular user which could benefit from this and we want to avoid that.

Feel free to argument here or to close the PR directly

denisab85:
Is there any valid reason for the plugin to be run with elevated rights? Just asking because I may overlook something that I don't know of. If there is no such a requirement, then it's 100% the admin's responsibility to do so and to deal with any security implications. My point of view, a tool that converts a set of text files into another set of text files should not require any special permissions. If the admin decided to give it such permissions, they are aware of the consequences. 
On the other hand, the error of "the folder does not exist" is something 99% of users will face when they start using the plugin. And this is a normal expectation that I believe sfdx-git-delta must satisfy.

scolladon:
I agree with you on the point that there should not be any reason for the plugin to run with elevated privilege.
But admin can install it using sudo, we cannot enforce they do not.
What we can do is to ensure the plugin does not create root output folder and only write in an existing location known by the user.
I agree with you it would be more convenient for the user if the plugin creates on the fly the output folder when it does not exist (fs-extra does it very well) but we decided to not do it long time ago for trust reason

mehdisfdc:
hi @denisab85 
First, thank you so much for taking the time to contribute and to propose enhancements to SGD, it's wonderful to see this level of involvement!

Regarding this specific feature, I tend to agree with @scolladon. SGD does not _need_ to create a new folder on the fly, and on the other hand it could in theory open to door to some "hard-to-exploit-but-not-impossible-anymore" scenarios were SGD would be used to send and write files in places it was not supposed to.

If SGD was meant to be used mainly manually, I'd say that it would be risk worth taking (because having the "output folder does not exist" is always annoying, even though it's very hard to fix). But my assumption is that most use-cases for SGD are within CI/CD pipeline, were the issue only needs to be fixed once for all (and again, this is very simple: `mkdir`). 

So at the end it's decision between a (very slightly) more secure tool, vs a (very slightly) easier-to-use tool. In that specific instance, I would lean towards staying with the "very slightly more secure tool" option.

But it's an interesting debate, and I'd love to hear if you don't agree with this view. And again thank you so much for your time in helping with SGD! We do not mean to discourage it, but simply to make sure the tool remain as lean and secure as possible :-)

denisab85:
@mehdisfdc, I see why it makes sense to stick with the more secure approach. I will accept your point of view here. Thanks!

PR #291, opened 2022-04-15
fix: link "Condition deployment" on TOC

tprouvot:

---
Fix the link on the table of content for the entry "Condition deployment on package.xml and destructiveChange content"


mehdisfdc (approved):
Thanks for the fix 👍

scolladon (approved):
LGTM

PR #322, opened 2022-08-02
feat: implement flow translation addition

### LINKED ISSUE #321: When a flow metadata is changed and the translations isn't, all the translations for the new flow version are lost
## What is the problem?

---

When we change a flow metadata and don't make any change to the translations file, after the deploy we loose all the translations related to that flow.
The reason for this is that the translation for each flow is managed by a flow version, so if a flow is deployed without the translation file, then there are no translations for its new version.

## Steps to reproduce

---

1. Have an env with a flow with translations
2. Use sgd to produce a package that detects a change only in the flow so a package.xml file is produced without the translations file
3. After the package is deployed, the new flow version will loose its translations

## Current workaround

---

Every time we touch a flow metadata, we make a small insignificant change to the translation file too so it will also be added to the package.xml

## Solution suggestion

If a flow was added to to the package.xml file and a translation wasn't, then force add translations.
This functionality might also be activated by a flag in the cli or some kind of configuration

a more advanced solution is to check which specific file translation contains a reference to that specific changed flow instead of adding the entire translations directory

scolladon:

---

- [X] Added (for new features).

## Explain your changes

---

Implement a new type handler for flow
When a flow is added or changed if it is referenced in a translation file then the translation file is added in the package and in the output directory as well

---

closes #321 

---

Use the branch [issue/321](https://github.com/scolladon/sfdx-git-delta-reproduction-playground/tree/issue/321) of the playground


mehdisfdc (approved):
Nice new feature!

PR #345, opened 2022-08-29
build: installation check with production context

scolladon:

---

## Explain your changes

---

Install the package as if it was in a production environment to simulate closely `sfdx plugins:install`
Goal is to avoid having install broken because of dependencies upgrade or development tool dependencies

**Other things:**
- use --frozen-lockfile to ensure no changes to the lock files is done during build in the env => meaning package.json and yarn.lock are aligned
- upgrade setup-node to v3
- remove fetch-depth full for publish job => no need
- remove cache for installation test => we want it to build fresh new
- install sfdx-cli using npm instead of yarn because the global yarn install is not reachable in windows build

---

Nop


mehdisfdc (approved):
LGTM :check-green:

PR #360, opened 2022-10-02
fix: ignore flow translation

### LINKED ISSUE #359: .forceignore failing to exclude "Translations" (latest working version: v5.3.1)
I observed today that translations are coming back to generated package.xml even though they are mentioned in .forceignore file, I have installed previous version v5.3.1 and in that version it's working as expected.
Kindly please fix that issue, maybe there are more metadata, but I only encountered this one

scolladon:

---

- [x] Fixed (for any bug fixes).

## Explain your changes

---

Create a `ignoreHelper` to be the adapter with the ignore module
It stores `ignore` instances per ignore file path (destructive and additive)
Make the `ignore` instances map static and use lazy loading to build it
Refactor repoGitDiff using the new ignoreHelper
Use the new helper in the `flowTranslationProcessor` module

---

closes #359 

- [x] Jest test added to check the fix is applied.


__tests__/unit/lib/utils/repoGitDiff.test.js
```diff
@@ -177,16 +177,16 @@ describe(`test if repoGitDiff`, () => {
     expect(work).toStrictEqual(expected)
   })
 
-  test('can filter deletion if only ignored is specified files', async () => {
+  test('cannot filter deletion if only ignored is specified files', async () => {
```
scolladon, line 180: **remarks:** I included this in the PR as a fix

I think the previous implementation where wrong in its spec, so I changed the test to be closer to what we need
And I think it should be like it is now.
I was not able to remember why it was like this and I assume it was a little mistake that slept through

I tested it functionally and it seems to work ok, it can be tested more

We can discuss it

mehdisfdc (approved):
LGTM

PR #382, opened 2022-11-03
feat: improve parameters usage description

scolladon:

---

## Explain your changes

---

Improve Admin experience by changing the name of the inputs, their descriptions and the possible value to put into

### DIFF
--- a/.github/workflows/manage-versions.yml
+++ b/.github/workflows/manage-versions.yml
@@ -11,8 +11,11 @@ on:
       version-alias:
         description: 'verion alias to map to the source version'
         required: true
-        default: 'stable | latest | latest-rc | vA.B.C'
-        type: string
+        type: choice
+        options: 
+        - stable
+        - latest
+        - latest-rc
 
 jobs:
   add-tag:
--- a/.github/workflows/manage-versions.yml
+++ b/.github/workflows/manage-versions.yml
@@ -3,15 +3,15 @@ name: Manage Versions
 on:
   workflow_dispatch:
     inputs:
-      source:
-        description: 'version number or alias to change from'
+      version-number:
+        description: 'source version number to use'
         required: true
-        default: 'latest-rc'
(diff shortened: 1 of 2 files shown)


.github/workflows/manage-versions.yml
```diff
@@ -3,15 +3,15 @@ name: Manage Versions
...
+      version-number:
+        description: 'source version number to use'
         required: true
-        default: 'latest-rc'
+        default: 'vX.Y.Z'
         type: string
-      target:
-        description: 'version number or alias to change to'
+      version-alias:
+        description: 'verion alias to map to the source version'
```
mehdisfdc, line 12:
```suggestion
        description: 'version alias to map to the source version'
```

.github/workflows/manage-versions.yml
```diff
@@ -3,15 +3,15 @@ name: Manage Versions
 on:
   workflow_dispatch:
     inputs:
-      source:
-        description: 'version number or alias to change from'
+      version-number:
+        description: 'source version number to use'
         required: true
-        default: 'latest-rc'
+        default: 'vX.Y.Z'
```
mehdisfdc, line 9: I think I preferred the previous default values.

When I see `latest-rc` and `latest`, I assume that it's proposing to promote `latest-rc` to `latest`, making it easy for me (as an admin) to understand the meaning of those 2 parameters.

On the other hand, if I see `vX.Y.Z` and `stable | latest | latest-rc | vA.B.C`, there is no relationship between those examples that would help me understand the meaning of the 2 parameters.

scolladon: Issue here is we cannot move an alias to another alias
We need to move an alias to a version number.
This is why I changed the default values, they works as a placeholder here.
We could use a dropdown list for the second one as it cannot be a version number and it is one of those 3 (stable latest or latest-rc)

mehdisfdc (approved):

PR #385, opened 2022-11-06
refactor: copy files using git instead of fs

### LINKED ISSUE #355: Allow to generate delat when "to" parameter is different than HEAD
Allow `-d` parameter to work with `-t` parameter different than HEAD

---

Issue #354 explained a clever usage in their pipeline where it could be very helpful.

---

Instead of using the file from the file system (current implementation), use the content of the file at the `--to` commit state when copying the incremental sources to the output folder.

`fsHelper.copyFiles` method needs to rely on `git show ${config.to}:${src)` content and use `fs-extra.outputFile` to write the file.
fsHelper unit test needs to be adapted as well as some integration test (not listed here)

---

None

### Additional context

---

issue #354

scolladon:

---

- [X] Added (for new features).
- [X] Fixed (for any bug fixes).

## Explain your changes

---

🎉 We now rely on git files instead of relying on the file system itself. 🎉

➕ It allows us to generate incremental package metadata event when `--to` is different than `HEAD`
➕It allows us to enforce `.gitignore` as well
➖ Also, it slows down a bit the performance, between 30% to 100% incremental package metadata generation time from my tests. No impact when generating just the xml files (`package.xml` and `destructiveChanges.xml`)

🚸 It contains a change in the `manage-version.yml` file so it is easier to understand the inputs
🚀 The wait for the installation time in the release workflow is now 30s instead of 300s
🚀 git hooks have been improved as well to be scoped to the content of the commit
🚧 Unit Test have been improved for the services. They are now more like specs and all the previous test have been moved (`test: add system integration testing` commit) to a `domain.test.js` file.
⛑ A few refactoring/simplification on the services have been applied during the process as well (look at the `refactor:` commits)
🧟‍♂️ Dead code has been removed as much as possible, not sure if I caught everything though.

---

closes #355

---

As usual, install the PR and play around with it.
Only new feature is to be able to generate delta with any pointer in the code:
```
$ git checkout feature/wip
$ sfdx sgd:source:delta -f "main" -t "release" -d 
```

## Any other comments

---

I have tested it locally with a huge project and it seems to work fine, performance aside
I have not play around with it much to look for corner cases
And I have not tested particular error cases as well
We may want to cover those before release it

[ADR log Link](https://github.com/scolladon/sfdx-git-delta/wiki/2022-11-11-Use-git-instead-of-file-system-to-generate-sources)


README.md
```diff
@@ -207,6 +207,11 @@ sfdx sgd:source:delta --from "HEAD~1" # right git shortcut with windows because
 
 ### CI/CD specificity
 
+In CI/CD pipelines, for most of the CI/CD provider, the checkout operation fetch only the last commit of the branch currently evaluated.
```
mehdisfdc, line 210:
```suggestion
In CI/CD pipelines, for most of the CI/CD providers, the checkout operation fetch only the last commit of the branch currently evaluated.
```

README.md
```diff
@@ -207,6 +207,11 @@ sfdx sgd:source:delta --from "HEAD~1" # right git shortcut with windows because
...
+This is done prbably to have the smallest network load possible and decrease clone time.
```
mehdisfdc, line 211: I would remove this sentence _("This is done probably to have the smallest network load possible and decrease clone time."_)

mehdisfdc (approved):
Nice feature! LGTM

mehdisfdc (approved):

PR #388, opened 2022-11-10
fix: ignore destructive with ignore when possible

### LINKED ISSUE #386: .sgdignore does not accept path exclusion, like !
Issue verification check:

## What is the problem?

---

I tried to "exclude all but some" and verified it does not work. In my .sgdignore file I had the following paths:

````
dir
!dir/dir
````

I this case Im expecting sgd to exclude all metadata changes from `dir` but the ones under `dir/dir` 

`sfdx sgd:source:delta -f <GIT_REF> -o <OUTPUT_FOLDER> -i <PATH_TO_SGDIGNORE>`

.sgdingore must exclude all metadata changes from `dir` but the ones under `dir/dir` 

.sgdingore with paths exclusions isn't working

## Steps to reproduce

---

1 - create a .sgdignore with that allows all files under a directory but exclude few using `!`

````
force/main/default/lwc
!force/main/default/lwc/cs_*
````

2 - run `sfdx sgd:source:delta -f <GIT_REF> -o <OUTPUT_FOLDER> -i <PATH_TO_SGDIGNORE>`
3 - verify that all lwc are included, and not just the ones prefixed with `cs_*`

## Execution context

---

image

scolladon:

---

- [X] Fixed (for any bug fixes).

## Explain your changes

---

Reapply business rule stating we should use the ignore pattern when `--ignore` is set and `--ignore-destructive` is not set.
Reapply the test that was previously validating this business rule
Refactor the service to be easier to unit test
Completely unit test the new ignore service (act as a spec)

Business rule is as follow:
| when                            | `--ignore` is set                                                                                            | `--ignore` is not set                                                                      |
|---------------------------------|-----------------------------------------------------------------------------------------------------------|-----------------------------------------------------------------------------------------|
| **`--ignore-destructive` is set**     | filter changed files with `--ignore` patterns filter deleted files with  `--ignore-destructive`  patterns | do not filter changed files filter deleted files with  `--ignore-destructive`  patterns |
| **`--ignore-destructive` is not set** | filter changed files with  `--ignore`  patterns filter deleted files with  `--ignore`  patterns           | do not filter changed files do not filter deleted files                                 |

---

closes #386

---

Play with ignore locally

### DIFF
--- a/__tests__/unit/lib/post-processor/flowTranslationProcessor.test.js
+++ b/__tests__/unit/lib/post-processor/flowTranslationProcessor.test.js
@@ -10,15 +10,11 @@ const {
   scanExtension,
   isSubDir,
 } = require('../../../../src/utils/fsHelper')
+const { forPath } = require('../../../../src/utils/ignoreHelper')
 jest.mock('fs-extra')
 jest.mock('fast-xml-parser')
 jest.mock('../../../../src/utils/fsHelper')
-const mockForPath = jest.fn()
-jest.mock('../../../../src/utils/ignoreHelper', () => {
-  return jest.fn().mockImplementation(() => {
-    return { forPath: mockForPath }
-  })
-})
+jest.mock('../../../../src/utils/ignoreHelper')
 
 const FR = 'fr'
 const EN = 'en'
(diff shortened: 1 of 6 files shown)


mehdisfdc (approved):
Thanks for the fix @scolladon !

PR #415, opened 2023-01-09
feat: silent flag parameter

### LINKED ISSUE #378: Clean output if json parameter is not set
### Clean output if json parameter is not set
Remove json output if --json parameter is not set, use spinner instead.

### What is the current result ?
`sfdx sgd:source:delta -s force-app -f <FROM> -t <TO> -o <OUTPUT>` 
**Result:**
`{
  "error": null,
  "output": "<OUTPUT>",
  "success": true,
  "warnings": []
}`

---

### What are the expected results ?

1. Without JSON parameter:
`sfdx sgd:source:delta -s force-app -f <FROM> -t <TO> -o <OUTPUT>` 
**Result:**
`Generating delta...done` or **nothing**

2. With JSON parameter
`sfdx sgd:source:delta -s force-app -f <FROM> -t <TO> -o <OUTPUT> --json` 
**Result:**
`{
  "error": null,
  "output": "<OUTPUT>",
  "success": true,
  "warnings": []
}`

gavignon:
Silent Flag to clean the output when --json parameter is not set and the command executed successfully

Resolves #378


CONTRIBUTING.md
```diff
@@ -11,6 +11,7 @@ We encourage the developer community to contribute to this repository. This guid
 
 - [Node](https://nodejs.org/) >= 14
 - [yarn](https://yarnpkg.com/) >= 1.22.5
+- [Code Climate](https://github.com/codeclimate/codeclimate) (If you are on MacOS, you will need Xcode Command Line Tools)
```
scolladon, line 14: **praises:** thanks, nice boy scout attitude 👍

messages/delta.js
```diff
@@ -17,4 +17,5 @@ module.exports = {
   includeFlag: 'file listing paths to explicitly include for any diff actions',
   includeDestructiveFlag:
     'file listing paths to explicitly include for any destructive actions',
+  silentFlag: 'no output if the command executed successfully, default to false.'
```
scolladon, line 20: **suggestion:** sentence simplification proposition

```suggestion
  silentFlag: 'does not display result in stdout/stderr'
```

src/commands/sgd/source/delta.ts
```diff
@@ -74,6 +74,10 @@ export default class SourceDeltaGenerate extends SfdxCommand {
       char: 'N',
       description: messages.getMessage('includeDestructiveFlag'),
     }),
+    'silent': flags.boolean({
+      char: 'S',
```
scolladon, line 78: **suggestion:** I think we should use 'i' instead of 'S'

Maj are used for parameter name with multiple words
As 's' is not available then we should use the next letter 'i'

src/commands/sgd/source/delta.ts
```diff
@@ -104,7 +108,9 @@ export default class SourceDeltaGenerate extends SfdxCommand {
       output.error = err.message
       process.exitCode = 1
     }
-    this.ux.log(JSON.stringify(output, null, 2))
+    if(output.success && !this.flags.silent){
+      this.ux.log(JSON.stringify(output, null, 2))
+    }
```
scolladon, line 113: **remarks:** I don't think we should gate the output with the success

In this case it means error will not be displayed anymore which is not what we want.

scolladon (changes requested):
I ask for a few changes (we can discuss them if you think otherwise)

I think we could add 2 test in `cli.test.js` to cover this new cli behavior

Look forward to read the next iteration, so far so good !

src/commands/sgd/source/delta.ts
scolladon: nevermind, 'i' is already used, 'S' is fine

PR #482, opened 2023-02-24
fix: `fsHelper.pathExist` when path does not exist

### LINKED ISSUE #481: Deleting one file from LWC bundle puts entire component to destructiveChanges
Issue verification check:

## What is the problem?

---

Deleting only one of the files in the LWC bundle (for example css file) leads to entire component being added to the destructiveChanges.xml

```
sfdx sgd:source:delta --to HEAD --from HEAD~1 --source="force-app/" --generate-delta --output __diff/ --ignore .forceignore
```

LWC component is added to the package.xml and diff sources folder.

LWC is added to the destructiveChanges.xml

## Steps to reproduce

- Clone the playground: https://github.com/nikitakarpenkov/sfdx-git-delta-reproduction-playground
- Run `sfdx sgd:source:delta -d -f HEAD~1`

## Execution context

---

**Operating System:** Ubuntu 22.04 and Mac OS Monterey 12.3

**yarn version:** 1.22.19

**node version:** 16.17.1

**git version:** 2.39.2

**sfdx version:** 7.188.1

**sgd plugin version:** 5.13.2

---

```bash
$ git diff --name-status --no-renames HEAD~1 HEAD
```

```
D       sgd/reproduction/playground/lwc/testComponent/testComponent.css
```

nikitakarpenkov:
## Explain your changes

Corrects a bug where an existing path in git was identified as non-existent.

---

closes #481

- [ ] Jest test added to check the fix is applied.

---

## Any other comments

---


scolladon:
Hi @nikitakarpenkov !

Thank you very much for this PR, really appreciated

I'll check the issue locally soon and the fix as well.

What would be great is to add a unit test in the related test class so we can make sure the failing use case is covered

src/utils/fsHelper.js
```diff
@@ -63,7 +63,7 @@ const readPathFromGit = async (path, config) => {
 
 const pathExists = async (path, config) => {
   const data = await readPathFromGit(path, config)
-  return data.startsWith(FATAL)
+  return !!data && !data.startsWith(FATAL)
```
scolladon, line 66: **suggestion:** I think an [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) would work here

```suggestion
  return !data?.startsWith(FATAL)
```

scolladon (approved):
LGTM

PR #506, opened 2023-03-10
build: script conveniency and improve documentation

AllanOricil:
## Explain your changes

- fixes wrong step number in the contribution guidelines.
- separate the build task from the pre prepackage hook. Now contributors can run `yarn build` instead of `rm -rf lib && tsc -b`

No

No

## Any other comments

No


scolladon (changes requested):
Very good idea and very clean work so far!
Thank you very much

I think the `.vscode/launch.json` should not be in the repo
I would prefer it to be inlined in the CONTRIBUTING.md
Then maybe we could also provide a way to [fill variables](https://code.visualstudio.com/docs/editor/variables-reference) instead of changing this `launch.json` manually with the parameters we want (which will be marked as changed by git if it stays in the repo).

What would suit me would be a reference to this [doc](https://code.visualstudio.com/docs/nodejs/nodejs-debugging) in the new section "Debugging with Visual Studio Code debugger" of the CONTRIBUTING.md

I like the new `build` command, it allows to not have to run the oclif commands and have something executable 👍

.vscode/launch.json
```diff
@@ -0,0 +1,23 @@
+{
+	"version": "0.2.0",
```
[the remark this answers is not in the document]
AllanOricil: I thought 0.2.0 was the schema version vscode would use to parse this json. Let me change that

AllanOricil:
@scolladon all debugging tasks are now configurable with input variables. Use `simple` if you only want to set the -f flag, and `full` if you want to specify values for all flags. I've also added this [link](https://code.visualstudio.com/docs/editor/debugging) in the contribution guidelines.

CONTRIBUTING.md
```diff
@@ -28,14 +28,58 @@ This will install all the tools needed to contribute
 yarn
 ```
 
-### 2) Build application
+### 3) Build application
```
scolladon, line 31: **praises:** nice "boy scout rule" attitude here, thanks

scolladon:
Thanks @AllanOricil that is very interesting.

On my side I have this kind of configuration:
```json
{
  "version": "0.2.0",
  "configurations": [
    {
      "type": "node",
      "request": "launch",
      "name": "Run selected Jest test",
      "runtimeExecutable": "npm",
      "runtimeArgs": [
        "run-script",
        "test"
      ],
      "args": [
        "--inspect",
        "--trace-warnings",
        "--",
        "-i",
        "${relativeFile}",
        "-t",
        "${selectedText}",
      ],
      "console": "integratedTerminal",
      "internalConsoleOptions": "neverOpen",
      "disableOptimisticBPs": true,
      "program": "${workspaceFolder}/node_modules/jest/bin/jest"
    }
  ]
}
```

I work with unit tests (TDD style) and so I really value being able to debug test in my IDE.
When I need to debug the app I run the following in the terminal:
```
$ node --inspect-brk bin/run sgd:source:delta <param list> <...> 
```

I really don't want the `.vscode/launch.json` file to be committed because I cherish the possibility for any community developer to work the way (s)he wants, whatever node / arch / system / keyboards / etc (s)he fancys
This is the reason why this file is in my `.git/info/exclude`
Screenshot 2023-03-10 at 18 17 58

Why it is not in the `.gitignore`? it is because I prefer when it is the developer who cleans the artifact related to its preferences.

What I would like to integrate in the codebase from this PR is the `build` script and the typo fix in the CONTRIBUTING.md

That being said it is very interesting what you managed to do with the parameters.
I strongly think it should be exposed to others via a ["Show & Tell" discussion](https://github.com/scolladon/sfdx-git-delta/discussions/categories/show-and-tell).

.vscode/launch.json
```diff
@@ -0,0 +1,242 @@
+{
```
scolladon, line 1: **praises:** impressive configuration 👍

scolladon (approved):
LGTM

PR #518, opened 2023-03-14
feat: add support for forecasting metadata types

jdkgabri:
## Explain your changes

Added forecasting related metadata types.
---

---

closes #

- [ ] Jest test added to check the fix is applied.

---

## Any other comments

Tested on Windows 11 with API version 57.

Not incrementing API version/release. To be decided up to the mantainers.

---

### DIFF
--- a/src/metadata/v57.json
+++ b/src/metadata/v57.json
@@ -7,11 +7,11 @@
     "xmlName": "ForecastingFilter"
   },
   {
-    "directoryName": "ForecastingFilterConditions",
+    "directoryName": "forecastingFilterConditions",
     "inFolder": false,
     "metaFile": false,
     "suffix": "forecastingFilterCondition",
-    "xmlName": "forecastingFilterCondition"
+    "xmlName": "ForecastingFilterCondition"
   },
   {
     "directoryName": "forecastingSourceDefinitions",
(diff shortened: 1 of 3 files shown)


src/metadata/v57.json
```diff
@@ -1,4 +1,39 @@
 [
+  {
+    "directoryName": "forecastingFilters",
+    "inFolder": false,
+    "metaFile": false,
+    "suffix": "forecastingFilter",
+    "xmlName": "ForecastingFilter"
+  },
+  {
+    "directoryName": "forecastingFilterConditions",
```
jdkgabri, line 10: Documentation states that folder and suffix start with Uppercase but it is wrong: https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_forecastingfiltercondition.htm

Proof:

image

src/metadata/v57.json
```diff
@@ -1,4 +1,39 @@
 [
+  {
```
scolladon, line 2: **remarks:** those metadata are available since v55

They should also be added in v55 and v56

scolladon (changes requested):
Thank you very much, nice job so far

I have one comment, please address it so we can include this contribution

src/metadata/v57.json
jdkgabri: All right. On it.

scolladon (approved):
LGTM

thanks for this contribution

PR #90, opened 2021-01-17
feat: add a new command to filter the content of Profile and Pset based on multiple package.xml

scolladon:
## What does this pull request contains? Explain your changes.

---

- [X] Added for new features.
- [ ] Changed for changes in existing functionality.
- [ ] Deprecated for soon-to-be removed features.
- [ ] Removed for now removed features.
- [X] Fixed for any bug fixes.
- [ ] Security in case of vulnerabilities.

## Explain your changes

---

Add a new command to have the ability to filter the content of a Profile and a Permission Set based on the content of multiple package.xml
The command takes a list of path of package.xml
The command takes a list of path where profile and permissionset should be located (recursively) to apply the filtering into them.
The command takes a list of userPermissions to keep. The actual behaviour is to remove every user permission not present in this white list.

## Does this close any currently open issues?

---

No issue related

- [ ] Jest test to check the fix is applied are added.

## Any particular element to being able to test locally

---

To test it you have to call the command `sfdx sgd:source:ppset -h`

## Any other comments?

---

This is WIP, no test case written yet and no traction from the community for this use case.
The current state of the code written is not acceptable

## Where has this been tested?

---

**Operating System:** Darwin Kernel Version 18.7.0

**Yarn version:** 1.22.10

**Node version:** v15.6.0

**sgd version:** 4.1.0

**git version:** 2.30.0

PR #533, opened 2023-03-26
feat: extend metadata content pruning to more eligible types

### LINKED ISSUE #198: Incrementally get Profile Changes

---

Similar to how it currently for custom label and workflows, We are not able to pick changes incrementally from Profile Files

---
So for example, there is a profile file, and I am newly adding a field permission to it. I would want want only the newly Field permission to be present while running the command.

---
Assume we have this profile file at commit abc123
image

And then we add a new field permission as follows at commit def456
image

After running the command
we want the file to look as follows

image

### Additional context

---
N/A

### LINKED ISSUE #328: [Metadata Files Scraping] Auto generation & scraping of profile, permission set & translation files

---

We often hit the limit of 10K lines with the profile & permission set metadata file. It is extremely hard to finetune those files as they require lots of manual work and advanced development skills.

---

Scrub the profile, permission set & translation metadata files based on recent changes, and or the package.xml file. Would be great to have --parameter for each scenario.

---

No current good-known alternative solutions.

### Additional context 

---

scolladon:
## Explain your changes

---

Extends metadata content filtering to those types:
- WorkFlow (and every sub types)
- SharingRule (and every sub types)
- AssignmentRule
- AutoResponseRule
- EscalationRule
- MatchingRule
- GlobalValueSetTranslation
- StandardValueSetTranslation
- Profile
- Translation
- CustomObjectTranslation

---

closes #198
closes #328
closes #525
closes #90 

- [x] Jest test added to check the fix is applied.


scolladon:
Question I have currently for this new feature: Should it be gated behind a cli parameter ? (like `--semantic-split` or `--smart-increment`)

mehdisfdc (dismissed):

mehdisfdc (approved):

PR #635, opened 2023-06-09
feat: add API v58 support

scolladon:
# Explain your changes

---

add API v58 metadata definition

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
   "author": "Sebastien Colladon <colladonsebastien@gmail.com>",
   "dependencies": {
     "@salesforce/command": "^5.3.9",
-    "fast-xml-parser": "^4.2.2",
+    "fast-xml-parser": "^4.2.4",
     "fs-extra": "^11.1.1",
     "ignore": "^5.2.4",
     "lodash": "^4.17.21",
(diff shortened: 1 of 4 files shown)


mehdisfdc (approved):

jonny-harte:
hi @scolladon when will this be the stable version?

scolladon:
Hi @jonny-harte, we'll move the latest and stable version to the latest version containing the v58 api today or tomorrow

Raspikabek:
Hey @scolladon , where are these json file definitions for metadata types coming from? I mean, is there any Salesforce documentation about this list of files? 

For context: I'm building a plugin myself and I need something similar to identify the metadata type name, directory file and if contains child elements like CustomObjects that contains CustomField.

(apologies this might not be the right place, but didn't see fair to raise an 'issue' for this.)

scolladon:
Hi @Raspikabek !

Those json files have been generated long time ago from the metadata wsdl if I remember correctly. We converted it from xml to json and then rework the format a bit so it is easier to consume.

But today there is another way to do it, much more easier. We have plan to refactor the plugin to use it.
It is the [Metadata Registry](https://github.com/forcedotcom/source-deploy-retrieve/blob/main/HANDBOOK.md#metadata-registry), if you need to have metadata definition in your plugin I think you should consider using this library.

I'll be happy to discuss design with you for your new plugin ! Contact me on [linkedin](https://www.linkedin.com/in/sebastien-colladon/)

PR #662, opened 2023-07-27
fix: stream in error without error content handling

### LINKED ISSUE #661: Intermittent erroneous empty files created in experiences and staticresources folders
We've just started using sfdx-git-delta - great tool, we're looking forward to using it on more projects.

Whilst testing, I found builds fail intermittently where the delta contains experience bundles or static resources. It appears that empty [something]-meta.xml files are created in random places in experiences/ and staticresources/ - these files don't and shouldn't exist.

This only happens on our CI builds (within a Docker container running on Bitbucket Pipelines) - but is reproducible in almost all builds containing these files. Though, I can't reproduce locally on my mac (and I've tried a lot...).

I've investigated the issue and it appears the tool attempts to get -meta.xml files from git that don't exist. I can see why that behaviour would exist - it's common that a static resource would change and therefore be included in the delta but the meta file wouldn't but both files need to be in the deploy. I think that logic could potentially be optimised in the future i.e. I don't think there's a need to check for -meta.xml files within sub folders of staticresources or sub folders of experiences, etc. However, that doesn't explain why the empty files are being created.

On further digging, I've traced the issue to the stream handling of processes - specifically the git show command used to retrieve file contents from the repository. What's happening is - there's an attempt to run git show for the file that doesn't exist but whilst the command is failing, the code is not waiting for that failure and instead proceeds to create the file (that doesn't exist) with no content (because git show correctly doesn't return anything).

In the stream handling there are two issues:
1. The code doesn't wait for the stream to close - it looks like it's attempting to wait for all of stdErr and stdOut but that code doesn't work (or doesn't always work) and that isn't the same as waiting for the stream to close
2. There is no check of the error code returned from the command - not all commands will output to stdErr and commands can output to stdErr without there actually being an error

This is the method in question:

```
const getStreamContent = async stream => {
  const content = []
  for await (const chunk of stream.stdout) {
    content.push(chunk)
  }
  const error = []
  for await (const chunk of stream.stderr) {
    error.push(chunk)
  }
  if (error.length > 0) {
    throw new Error(error.join(''))
  }
  return Buffer.concat(content)
}
```
in src/utils/childProcessUtils.js

Async/await syntax appears to be quite problematic with streams, instead I propose this:

```
const getStreamContent = async stream => {
  return new Promise((resolve, reject) => {
    const content = []
    const error = []

    stream.stdout.on('data', data => {
      content.push(data)
    })

    stream.stderr.setEncoding('utf8')
    stream.stderr.on('data', data => {
      error.push(data.toString())
    })

    stream.on('close', code => {
      if (code != 0) {
        reject(new Error(error.join('')))
      }

      resolve(Buffer.concat(content))
    })
  })
}
```

Listen for data on stdout and stderr as before but only return when the stream has closed. Determine success or failure based on the process error code. And wrap the whole thing in a promise.

This is a like for like replacement of what's already there whilst resolving the bug.

I've applied this fix to our build and no longer see any issues in Bitbucket Pipelines.

I'm committing the changes at the moment and a PR will be raised shortly.

Thanks again for your work on this tool - it's awesome.

gdman:
# Explain your changes

---

Changes as follows:
- Replace stream processing with Promise syntax waiting for stream close
- Determine success or failure based on exit code
- Update tests to only emit close once and delay emitting slightly to make the tests more realistic

See #661 for full details.

---

closes #661

- [x] Jest tests added to cover the fix. - test updated and new test added for exit code
- [ ] NUT tests added to cover the fix. - n/a - internal implementation change
- [ ] E2E tests added to cover the fix. - n/a - internal implementation change


gdman:
npm-lint outdated dependencies fixed

__tests__/unit/lib/utils/childProcessUtils.test.js
```diff
@@ -70,6 +71,36 @@ describe('childProcessUtils', () => {
         }
       })
     })
+
+    describe('when stream emits error but no stderr output', () => {
+      it('throws an empty error', async () => {
+        // Arrange
+        expect.assertions(1)
+        const stream = new EventEmitter()
```
scolladon, line 79: **thought:** I wonder if it could be easy to isolate the generation of the stream in a test helper method

That way it would be easier to read the arrange part of those test.
Not sure it is worth it though as this part of the code may not change soon and is subject to be replaced by a future refactoring.

Something like this could be nice: 
```js
const arrangeStream = (data, error, isError) => {
  const getReadable = content => {
    return new Readable({
      read() {
        if (content) this.push(content)
        this.push(null)
      },
    })
  }
  const stream = new EventEmitter()
  stream.stdout = getReadable(data)
  stream.stderr = getReadable(error)
  setTimeout(() => stream.emit('close', isError ? 1 : 0), 0)
  return stream
}
```

then we could write line 79
```js
const stream = arrangeStream('irrelevant std out output', null, true)
```

__tests__/unit/lib/utils/childProcessUtils.test.js
```diff
@@ -40,26 +40,27 @@ describe('childProcessUtils', () => {
       }
     )
 
-    describe('when stream emit error', () => {
+    describe('when stream emits error', () => {
```
scolladon, line 43: **praises:** thanks for fixing the typo

src/utils/childProcessUtils.js
```diff
@@ -34,18 +34,27 @@ async function* linify(stream) {
...
+    })
+
+    stream.on('close', code => {
+      if (code != 0) {
+        reject(new Error(error.join('')))
+      }
+
+      resolve(Buffer.concat(content))
+    })
+  })
```
scolladon, line 57: **praises:** nit implementation

__tests__/unit/lib/utils/childProcessUtils.test.js
```diff
@@ -70,6 +71,36 @@ describe('childProcessUtils', () => {
         }
       })
     })
+
+    describe('when stream emits error but no stderr output', () => {
```
scolladon, line 75: **praise:** great new unit test to isolate the issue !

scolladon (commented):
Great job

I have a few question / thought but nothing blocking, the quality is great, thank you very much
I'll approve once we close every discussions.

~~In order to resolve the linting issue you can run the `upgrade:dependencies` script with `yarn`~~

__tests__/unit/lib/utils/childProcessUtils.test.js
```diff
@@ -70,6 +71,36 @@ describe('childProcessUtils', () => {
...
+          read() {
+            this.push('irrelevant std out output')
+            this.push(null)
+          },
+        })
+
+        stream.stderr = new Readable({
+          read() {
+            this.push(null)
+            stream.emit('close', 1)
```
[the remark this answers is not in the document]
gdman: I noticed the same thing, fixed when I updated the dependencies

__tests__/unit/lib/utils/childProcessUtils.test.js
gdman: Yes, very good idea

__tests__/unit/lib/utils/childProcessUtils.test.js
```diff
@@ -21,16 +21,16 @@ describe('childProcessUtils', () => {
...
             },
           })
           stream.stderr = new Readable({
             read() {
               this.push(null)
-              stream.emit('close')
             },
           })
 
+          setTimeout(() => stream.emit('close', 0), 10)
```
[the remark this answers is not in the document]
gdman: Ah, I set it to 10 to verify the issue before I fixed and I forgot to change it back. Updated to zero and the tests still pass 👍🏻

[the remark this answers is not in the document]
scolladon: Ok got it, I thought it was about some performances trick in IE I read about :D

gdman:
Code updated to resolve PR feedback, thanks for reviewing so quickly!

__mocks__/child_process.js
```diff
@@ -26,9 +25,9 @@ childProcess.spawn.mockImplementation(() => {
         this.push('error')
       }
       this.push(null)
-      mock.emit('close')
     },
   })
+  setTimeout(() => mock.emit('close', error ? 1 : 0), 10)
```
scolladon, line 30: **suggestion:** we could also apply 0ms here ?

```suggestion
  setTimeout(() => mock.emit('close', error ? 1 : 0), 0)
```

src/utils/childProcessUtils.js
```diff
@@ -34,18 +34,27 @@ async function* linify(stream) {
...
+      content.push(data)
+    })
+
+    stream.stderr.setEncoding('utf8')
+    stream.stderr.on('data', data => {
+      error.push(data.toString())
+    })
+
+    stream.on('close', code => {
+      if (code != 0) {
```
scolladon, line 51: **suggestion:** we may have to check with `!==``

```suggestion
      if (code !== 0) {
```

I think it is the last one, sorry for the late notice

scolladon (approved):
LGTM

Great contribution here, thank you very much

PR #702, opened 2023-10-04
fix: use default export for main module

### LINKED ISSUE #701: Changed export from default to named
We are using this tool as an imported module in our deployment tool. It works perfectly. Unfortunately we noticed that along with the latest v5.26.0 some breaking change has been introduced. Previously (i.e. v5.25.2 and older) the main function was exported as a default export. Then in order to consume the following code was required:

```javascript
const sgd = require('sfdx-git-delta')
await sgd({params})
```

That way is still in [readme](https://github.com/scolladon/sfdx-git-delta/tree/v5.26.0#use-the-module-in-your-own-node-application), however the behaviour has changed to named export, and now it requires:

```javascript
const { sgd } = require('sfdx-git-delta')
await sgd({params})
```

It would be great to follow semantic versioning rules and not to introduce breaking changes with minor changes. In other words please keep default export when doing minor version bumps.

scolladon:
# Explain your changes

---

Use default export instead of named export.

This [commit](https://github.com/scolladon/sfdx-git-delta/commit/443034b31bae3b07bf450e53eff506989e91d620) introduced a breaking change in the way the main module is exported.
It used a named export instead of a default export.
This change was introduce inside a minor version with a breaking change which is not compliant with the semantic versioning rules.
The change aim to fix that

---

closes #701


scolladon:
The megalinter lint issue seems to be an issue with trivy
I have not dug deeper

The npm-lint issue is something I could not fix with intermediate version or lock file fixing. We need to wait the next release of `swc/core`

mehdisfdc (approved):

PR #705, opened 2023-10-05
WIP: fix: restore default export

pawel-id:
# Explain your changes

---

It seems typescript clause `export default someting` is translated into `exports.default = sgd;` which is somehow different that expected `module.exports = sgd`. Then having this approach in order to import this we need to do some extra work:

```
const sgd = require('module').default
```

or 

```
import sgdDefault from 'module'
const sgd = sgdDefault.default
```

This is some proposal of solving the issue which I found. It is rather provided as example, not necessarily to merge (maybe there is smarter way).

related to https://github.com/scolladon/sfdx-git-delta/issues/701

### DIFF
--- a/src/index.cjs
+++ b/src/index.cjs
@@ -1,0 +1,2 @@
+const sgd = require('./main.js').default
+module.exports = sgd
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,5 +1,5 @@
 {
-  "extends": "./node_modules/@salesforce/dev-config/tsconfig",
+  "extends": "./node_modules/@salesforce/dev-config/tsconfig.json",
   "compilerOptions": {
     "declaration": true,
     "outDir": "./lib",
@@ -28,6 +28,6 @@
   },
   "include": [
     "./src/**/*",
-    "./src/**/*.json"
+    "./src/**/*.json",
   ]
 }
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "/messages",
     "/oclif.manifest.json"
   ],
-  "main": "lib/main.js",
+  "main": "lib/index.cjs",
   "repository": {
     "type": "git",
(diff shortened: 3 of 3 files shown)


pawel-id:
There something wrong in the code above, but I can not fix this by pushing new changes because test:nut failng. This is probably my fault with project configuration. Will close this PR.

tsconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-  "extends": "./node_modules/@salesforce/dev-config/tsconfig",
+  "extends": "./node_modules/@salesforce/dev-config/tsconfig.json",
```
scolladon, line 2: **question:** why should we add the extension here?

Is it solving something ?

src/index.cjs
```diff
@@ -0,0 +1,2 @@
+const sgd = require('./main.js').default
+module.exports = sgd
```
scolladon, line 2: **praises:** Smart way to workaround

scolladon (commented):
Thanks for this solution

I think it could work if we do
```js
export = sgd
```
instead of 
```js
export default sgd
```
line 30 in `main.ts`

Then we won't need to have the `index.cjs` file
What do you think @pawel-id, could you test it on your side ?

tsconfig.json
pawel-id: tsconfig.json change shouldn't be there. Please ignore it.

checking the solution with `export = sgd`...

pawel-id:
> Thanks for this solution
> 
> I think it [could work](https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require) if we do
> 
> ```js
> export = sgd
> ```
> 
> instead of
> 
> ```js
> export default sgd
> ```
> 
> line 30 in `main.ts`
> 
> Then we won't need to have the `index.cjs` file What do you think @pawel-id, could you test it on your side ?

@scolladon This works and it is super simple!

PR #706, opened 2023-10-06
fix: typescript default exported value

### LINKED ISSUE #701: Changed export from default to named

scolladon:
# Explain your changes

---

Fix how the default is exported in typescript
Make sure the compilation produce code that can be called those ways:
```js
const sgd = require('sgd')
import sgd from 'sgd'
```

`require` contract is tested in a dedicated `functional` test
`import` contract is tested in the `nut` tests and `e2e` tests

---

closes #701 


scolladon:
Could you help us with test @pawel-id please ?

Here are to [step to install a PR locally](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request)

pawel-id:
Hi @scolladon,

I did check this version as module referenced in our deployment tool and this time everything is fine. The behavior is exactly as expected: sfdx-git-delta exports single function sgd.

I also [installed plugin locally](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) and did few checks on our repository. They were fine too.

It looks good to me.

scolladon:
Thank you very much for your contribution!
It was a bumpy road 😅, I learned a lot, thank you for that too

mehdisfdc (approved):

Issue #717, opened 2023-11-01
Package.xml only has existing modified files, missing newly created files


MattFaz:
Issue verification check:

- [ ] is the current repository fully deployable at the commit SHA provided with the 'from' parameter of the command?

When I create a PR from `fromBranch` to `toBranch` I can see over 1k commits and 2k files changed. Not sure if there is any limits to sfdx-git-delta?

## What is the problem?

We are testing changing our CICD process to include sfdx-git-delta and facing issues with a large PR not including newly created files, only modified files.

### What is the parameter and the value you used with it?

In this testing the `fromBranch` and `toBranch` are being provided by the user.

`sf sgd:source:delta --from origin/${{ github.event.inputs.fromBranch}} --to origin/${{ github.event.inputs.toBranch }} --output .`

### What is the expected result?

Expect it to create a package.xml file with all created and modified files listed.
As an example I will refer to flows _(although it happens across all objects/created files)_, we can see there is 2 modified flows, and n newly created files.

![image](https://github.com/scolladon/sfdx-git-delta/assets/10377148/45e7d6ee-705c-477c-8dc0-97fd7bc3f899)

### What is the actual result?

It is only adding modified files to package.xml, not newly created files.

```
<types>
    <members>Account_Create_DSR</members>
    <members>AM_Handoff</members>
    <name>Flow</name>
</types>
```

## Steps to reproduce

I'm not entirely sure how to share a reproduction as I am unable to share the code due to workplace security. Full Github Workflow below.

## Execution context

OS: ubuntu-latest
Node v18.18.2
Git: 2.42.0
SFDX: @salesforce/cli/2.14.6 linux-arm64 node-v18.18.2
sfdx-git-delta 5.27.0 (latest-rc)

## More information (optional)

This is a very large difference, over 1k commits a 2k modified files.

Running `git diff --name-status --no-renames origin/fromBranch origin/toBranch` correctly lists all files.

Git Workflow below.

```
name: Manual Deployment

on:
    workflow_dispatch:
        inputs:
            fromBranch:
                description: 'FROM Branch'
                required: true
            toBranch:
                description: 'TO Branch'
                required: true
            toOrg:
                description: 'TO Org'
                required: true

jobs:
    validate:
        runs-on: ubuntu-latest
        steps:
            - name: 'Checkout source code'
              uses: actions/checkout@v4
              with:
                fetch-depth: 0

            - name: 'Determine org'
              run: |
[...]

PR #718, opened 2023-11-03
docs: clarify `from` and `to` parameters

scolladon:
# Explain your changes

---

Clarify usage of `from` and `to` parameters
Avoid confusion like in #717


mehdisfdc (approved):
It's clearer this way, thank you!

PR #731, opened 2023-12-04
feat: adding prediction builder ai metadata types

gaviriajesus:
# Explain your changes

Adding 4 metadata types to the different metadata files to allow Einstein Prediction Builder delta deployments.

---

No Issue related to this.

---

# Any other comments

an AI Model was created and deployed through VS Code with these files and it works so deltas will work as well, the  

directoryName, suffix and xmlName were triple checked with a local example.
inFolder and metaFile where set to false as I copied the connectedApps one.
---

### DIFF
--- a/src/metadata/v50.json
+++ b/src/metadata/v50.json
@@ -1798,31 +1798,31 @@
     "excluded": true
   },
   {
-      "directoryName": "aiApplicationConfigs",
-      "inFolder": false,
-      "metaFile": false,
-      "suffix": "aiapplicationconfig",
-      "xmlName": "AIApplicationConfig"
-    },
-    {
-      "directoryName": "aiApplications",
-      "inFolder": false,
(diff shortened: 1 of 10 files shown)


scolladon (changes requested):
LGTM

Nice addition, thank you. 
I checked API version, xml names, directory, suffix and other particularities and it seems fine !

Could you just add those metadata definitions at the beginning of the file please @gaviriajesus ? The end of the file is used for other special metadata currently

gaviriajesus:
> LGTM
> 
> Nice addition, thank you. I checked API version, xml names, directory, suffix and other particularities and it seems fine !
> 
> Could you just add those metadata definitions at the beginning of the file please @gaviriajesus ? The end of the file is used for other special metadata currently

On it.

scolladon (approved):
LGTM

Issue #350, opened 2022-09-14
Apex classes outside of force-app/main/default/classes not being added to package.xml


fedegiust:
Issue verification check:

- [x] is the current repository fully deployable at the commit SHA provided with the 'from' parameter of the command?

## What is the problem?

Currently we moved the Test Apex Classes to force-app/tests and have all other classes in the standard force-app/main/default/classes

If we run a git diff command, I can see apex classes and test classes in the modified files

![image](https://user-images.githubusercontent.com/1804862/190156236-a108d12a-c642-45d5-80fa-64f47b5bb7d7.png)

This is while sitting in pre-master branch to get the diff between pre-master and master to be able to run validation before merging into master

I tried the following commands
`sfdx sgd:source:delta --from $(git merge-base pre-master origin/master)  --output "./temp/" -s . -n .includeFolders`
`sfdx sgd:source:delta --from pre-master  --output "./temp/" -s . -n .includeFolders`
`sfdx sgd:source:delta --from origin/master  --output "./temp/" -s . -n .includeFolders`

The test classes never get added

### What is the parameter and the value you used with it?

`sfdx sgd:source:delta --from $(git merge-base pre-master origin/master)  --output "./temp/" -s . -n .includeFolders`
`sfdx sgd:source:delta --from pre-master  --output "./temp/" -s . -n .includeFolders`
`sfdx sgd:source:delta --from origin/master  --output "./temp/" -s . -n .includeFolders`

This is the content of my .includeFolders

```
force-app/main/default/
force-app/tests/
```

### What is the expected result?

```
<?xml version="1.0" encoding="UTF-8"?>
<Package xmlns="http://soap.sforce.com/2006/04/metadata">
    <types>
        <members>LeadsService</members>
        <members>LeadTriggerHandler</members>
        <members>LeadTriggerHandlerTest</members>
        <members>TestDataFactory</members>
        <name>ApexClass</name>
    </types>
    <types>
        <members>OpportunityRenewalsNew</members>
        <name>ApexPage</name>
    </types>
    <types>
        <members>Contact.Person_Source__c</members>
        <members>Lead.Person_Source__c</members>
        <members>Lead.TeamWhoCreated__c</members>
        <name>CustomField</name>
    </types>
    <types>
        <members>Account-Project Manager</members>
        <members>Contact-Channel Sales</members>
        <members>Contact-Internal Sales</members>
        <members>Contact-Sales</members>
[...]

PR #740, opened 2023-12-13
refactor: create git adapter

scolladon:
# Explain your changes

---

Create a git adapter layer
Implement most of the git methods with `isomorphic-git` and some that were not available with `simple-git`

PR #750, opened 2024-01-03
feat: resolve metadata type using file extension

### LINKED ISSUE #642: Detect metadata type based on file extension instead of directory

Metadata is not picked up and handling is too opaque. I had an issue where my directory name for PlatformEventSubscriberConfigs had a lower-case P at the start e.g. platformEventSubscriberConfigs. SFDX handles this fine but sgd does not. 

I was also experimenting with putting my test classes in a test directory. But this resulted in sgd not picking up changes to those files. 

---

Ignore the directory structure when examining changes and if there is change that is not matched to metadata within project directories they should be output as a warning in the output. 

---

If a change is detected only the extension is used to identify the metadata type. 

---

none.

### Additional context

---

scolladon:
# Explain your changes

---

Allow to resolve metadata type using the file extension instead of the folder when possible.

Fallback is to use the folder extension for those types:
- LightningWebComponent
- StaticResource

---

closes #642 #350 


scolladon:
Hi @fedegiust, @jverelst & @jjulicher !

We have a first working implementation of the metadata detection based on the extension first !

Could you please help us in validating this feature ?

If you are interested in the spec it is written [here](https://github.com/scolladon/sfdx-git-delta/pull/750/files#diff-e1c397f4850f2bc5c1b63aed5dfb487f7952fa5b1c93aac3fcd2b23513f423f8R1).
If you think we can add more cases feel free to suggest !

The code currently handles all the special cases we have found so far. Even the one when we have the [same extension but not the same metadata](https://github.com/scolladon/sfdx-git-delta/pull/750/files#diff-2edd1cc79f814187f5716f5ce0f60b44dceda0821c10af65bcfb14fed9a4735aR53) type because it is not contained in the same folder (one of the most tricky to make work with all the other cases)

In order to test it locally, you can follow those [steps](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) and install this PR.

Let us know, we will merge it soon and ship a new release asap

mehdisfdc (approved):
@scolladon As discussed, let's include this one (and #740 ) in a release candidate, to make it easier to test it extensively 😃

PR #885, opened 2024-06-14
fix: use `keywords` instead of `keyword` in package.json

amtrack:
Thanks for this awesome plugin. I'm using it in several projects!

I'm working on a website for exploring awesome sfdx/sf plugins and was wondering why it didn't show `sfdx-git-delta`.

https://amtrack.github.io/sf-plugin-explorer/

Untitled

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,7 +2,7 @@
   "name": "sfdx-git-delta",
   "version": "5.40.2",
   "description": "Generate the sfdx content in source format and destructive change from two git commits",
-  "keyword": [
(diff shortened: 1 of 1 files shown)


scolladon (approved):
LGTM !

Thanks for this contribution in the plugin visibility

scolladon:
I changed the title to "build" instead of "fix" to avoid creating a release candidat PR and avoid having this listed in the release note as it is not a feature nor a fix for the plugin itself

Issue #366, opened 2022-10-11
Deployment fails when deleting a flow


jdgirald:
Issue verification check:

- [x] is the current repository fully deployable at the commit SHA provided with the 'from' parameter of the command?

## What is the problem?

I deleted a flow from our repo (Account_ACU_Update) and then we ran the command `sfdx sgd:source:delta -f $MERGE_BASE -s src --output output` and it created a destructiveChanges.xml with a flow element. 

```
<?xml version="1.0" encoding="UTF-8"?>
<Package xmlns="http://soap.sforce.com/2006/04/metadata">
    <types>
        <members>Account_ACU_Update</members>
        <name>Flow</name>
    </types>
    <version>54.0</version>
</Package>
```

If I try to deploy it with `sfdx force:source:deploy --checkonly -x output/package/package.xml --postdestructivechanges=output/destructiveChanges/destructiveChanges.xml ` I get the following error:

Error   Account_ACU_Update insufficient access rights on cross-reference id

### What is the expected result?

Flow deleted from org or not added to destructiveChanges.xml

### What is the actual result?

Error   Account_ACU_Update insufficient access rights on cross-reference id

scolladon:
Hi @jdgirald 

Thanks for raising this issue!

I think it is related to this [Salesforce issue](https://trailblazer.salesforce.com/issues_view?id=a1p4V000002KI7bQAG&title=support-deletion-of-flow-metadata-through-destructive-changes)

The solution/workaround they are suggesting is to add the flow version in the `destructiveChanges.xml` instead
Is it possible in your workflow/repo to have the version in the flow file name when adding them in your repository ?

jdgirald:
Thanks @scolladon, looks like is a salesforce limitation.

Since we use the destructiveChanges.xml in automatic deployments, I don't think is possible to add the flow version to the file manually. We could somehow query the version number, but still, I guess the flow version might be different between environments.

For now, I think the workaround is to add flows to the ignore-destructive file, so they are not added to the `destructiveChanges.xml`.

scolladon:
I see it is [not possible anymore](https://docs.copado.com/articles/#!copado-ci-cd-publication/how-to-deploy-flows-using-salesforce-api-version-v-44-0) to retrieve flow file with their version number.

I was using this trick in a previous project: retrieving flow with their version number and deploying them with their version.
I solved our problem at that time which was to know which version is enabled.
But it had also some downside: 

[...]

PR #893, opened 2024-07-17
feat: deactivate Flow instead of trying to delete

### LINKED ISSUE #588: Create FlowDefinitions for deleted Flows

I'd like to be able to leverage SGD to delete flows, but due to a [Salesforce limitation](https://issues.salesforce.com/issue/a028c00000gAwixAAC/deletion-of-flow-metadata-through-destructive-changes-not-supported) ("Working as Expected"), I doubt that's ever going to be the case.

While `Flow`s can't easily be deleted in `destructiveChanges.xml`, we _can_ deactivate flows by deploying a `FlowDefinition` with `activeVersionNumber` set to `0`. We can later purge inactive flows.

e.g.

`My_Flow.flow-meta.xml`
```xml
<?xml version ="1.0" encoding="UTF-8"?>
<FlowDefinition xmlns="http://soap.sforce.com/2006/04/metadata">
    <activeVersionNumber>0</activeVersionNumber>
</FlowDefinition>
```

I propose adding functionality to SGD that would allow for creation of these `FlowDefinition`s. Given how this would work, it'd likely have to be behind an additional flag (e.g. `--transform-destructive-flows`), and would only work with `--generate-delta` (the README would have to make this very clear, and use of the new flag without `--generate-delta` should probably throw an error.

I've considered writing a script to do this, which wouldn't be complicated, but I think this fits in within the goals of SGD.

### Additional context

Earlier related issue: #366
Salesforce Known Issue: https://issues.salesforce.com/issue/a028c00000gAwixAAC/deletion-of-flow-metadata-through-destructive-changes-not-supported

amtrack:
# Explain your changes

When a Flow is detected in `destructiveChanges.xml` and the `--generate-delta` flag is given

- remove the Flow from destructiveChanges.xml
- add the Flow as FlowDefinition to package.xml
- generate a FlowDefinition file with `activeVersionNumber=0` to deactivate it

---

---

closes #588, #366

- [ ] Jest tests added to cover the fix.

---

# Any other comments

---


src/post-processor/deactivateFlowProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+            store: this.work.diffs.destructiveChanges,
+            type: FLOW_XML_NAME,
+            member: flowName,
+          })
+          fillPackageWithParameter({
+            store: this.work.diffs.package,
+            type: FLOW_DEFINITION_TYPE,
+            member: flowName,
+          })
+          await writeFile(
```
scolladon, line 56: **praises:** nice usage of the adapter

src/post-processor/deactivateFlowProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+            join(
```
scolladon, line 57: **remarks:** it should include the output as base

src/post-processor/deactivateFlowProcessor.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export default class DeactivateFlowProcessor extends BaseProcessor {
+  constructor(work: Work, metadata: MetadataRepository) {
+    super(work, metadata)
+  }
+
+  public override async process() {
+    if (this._shouldProcess()) {
+      const deletedFlows = this.work.diffs.destructiveChanges.get(FLOW_XML_NAME)
+      if (deletedFlows) {
```
scolladon, line 40: **suggestion:** it could be done in the `_shouldProcess` method

src/post-processor/postProcessorManager.ts
```diff
@@ -3,11 +3,13 @@ import { MetadataRepository } from '../metadata/MetadataRepository'
 import type { Work } from '../types/work'
 
 import BaseProcessor from './baseProcessor'
+import DeactivateFlowProcessor from './deactivateFlowProcessor'
 import FlowTranslationProcessor from './flowTranslationProcessor'
 import IncludeProcessor from './includeProcessor'
 import PackageGenerator from './packageGenerator'
 
 const processors: Array<typeof BaseProcessor> = [
+  DeactivateFlowProcessor,
```
scolladon, line 12: **question:** I wonder if it should be done after the FlowTranslation 🤔 ?

In this case we don't care about translations right ?

src/utils/packageHelper.ts
```diff
@@ -53,3 +53,18 @@ export const fillPackageWithParameter = ({
...
+  type,
+  member,
+}: {
+  store: Manifest
+  type: string
+  member: string
+}) => {
+  if (!store.has(type)) {
+    store.set(type, new Set())
+  }
```
scolladon, line 68: **remarks:** this code can be removed IMHO

src/utils/packageHelper.ts
```diff
@@ -53,3 +53,18 @@ export const fillPackageWithParameter = ({
   }
   store.get(type)?.add(member)
 }
+
+export const removeComponentFromPackage = ({
```
scolladon, line 57: **remarks:** this method should be tested :)

scolladon (commented):
Very promising start !

There are a few thing I've noticed
And also the test needs to be completed for the PR to be ready for review
I'll leave my comments for now.

Do not invest more time for now, I'm thinking about this feature and I'm not sure we want it or we want it this way

mehdicherf:
@scolladon @amtrack Sorry if this is a stupid, but what issue is this PR solving?

@amtrack Did you face any problem with the behavior of current plugin when it comes to delete flows?

scolladon:
Hi @amtrack,

Thanks for your work here, it made the topic hot again !

We discussed this with @mehdicherf and we tried to think of how it would change the usage of sgd and how it would force or allow users to have their own workflow.
What came out is that we prefer to not force the user to follow this workflow as it based on FlowDefinition (v44 or below) and we prefer users to not follow any workflow and prefer allowing anyone to build their own using the tool.

What we think could be done in this PR is to just add a warning inside the `work` object (`work.warnings.push` basically) just to let the user he is trying to add Flow into a `destructiveChanges.xml and it will probably not do what he expect. It could use the content [here](https://github.com/scolladon/sfdx-git-delta/issues/588#issuecomment-1534318729) and create a section in the README the message can point to.

So you could create another service extending standardHandler, just for Flow. and override the `handleDeletion` to add the warn while calling `super.handleDeletion`.
I can take care of creating this new PR, make the change in the README and add you as a reviewer if you like ?

amtrack:
@scolladon OK, I understand this decision.
Having a warning message printed asking the users to upvote the [Salesforce Known Issue](https://issues.salesforce.com/issue/a028c00000gAwixAAC/deletion-of-flow-metadata-through-destructive-changes-not-supported) and pointing out some workarounds sounds good to me.
Yes, please take over implementing the warning message.

PR #896, opened 2024-07-30
fix: issue copying resource, meta and folder for resource type

### LINKED ISSUE #895: Resource file changed does not copy the whole static resource
Issue verification check:

## What is the problem?

sfdx git delta plugin is not picking the source file in the delta directory if the change is done in the meta xml of a static resource.

Retrieve the static resource from the org in source format using the sf project retrieve start command
and then make a change to the -meta.xml file. Run the sfdx-git delta command to generate the delta folder.
Check the contents of the foler,it only has the -meta.xml file

@salesforce/cli/2.51.6 linux-x64 node-v18.17.1

@salesforce/cli: Installing plugin sfdx-git-delta@latest... installed v5.40.2

scolladon:
# Explain your changes

---

Ensure we copy every files for resource
- The resource files
- The metadata file

Let the fsHelper layer handle duplicate call to copyFiles

---

closes #895


scolladon:
Hi @RohitGagan16 !

Could you test on your side and help us validate the fix please ?
You can follow those [steps](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) to easily do it on your side

mehdicherf (dismissed):

PR #901, opened 2024-08-15
fix: suffix based translation file detection and output subDir

### LINKED ISSUE #321: When a flow metadata is changed and the translations isn't, all the translations for the new flow version are lost

scolladon:
# Explain your changes

---

Ensure the flow translation code deals with both `.translation` and `.translation-meta.xml` Translation files suffix.
Test if the translation file is a sub dir of the output folder when output folder is different than the repository folder

---

closes #321 


scolladon:
Hi @ArielS1 !

Could you test on your side and help us validate this PR please ?
You can follow those [steps](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) to easily do it on your side

ArielS1:
actually the file's extension on the real repo translation-meta.xml
but i'm still able to reproduce it both on dev-901 and 5.42.0
check my repository

scolladon:
> actually the file's extension on the real repo translation-meta.xml but i'm still able to reproduce it both on dev-901 and 5.42.0 check my repository

on my side it is fixed with the version in this PR...
I mean the translation file is added in the package.xml and is is also copied in the output folder.

Are you sure you are using the version of the plugin from this PR ?

ArielS1:
> > actually the file's extension on the real repo translation-meta.xml but i'm still able to reproduce it both on dev-901 and 5.42.0 check my repository
> 
> on my side it is fixed with the version in this PR... I mean the translation file is added in the package.xml and is is also copied in the output folder.
> 
> Are you sure you are using the version of the plugin from this PR ?

did you pull my last changes in my repo?
this is what i do, and the generated package.xml is like in the repo.

arielsi@Ariel:/mnt/c/Users/arielsi/IdeaProjects/SgdFlowTranslations$ sfdx plugins:install sfdx-git-delta@dev-901
 ›   Warning: sfdx-cli update available from 7.172.0 to 7.209.6.
This plugin is not digitally signed and its authenticity cannot be verified. Continue installation y/n?: y
Finished digital signature check.
warning "sfdx-git-delta > @salesforce/command > @oclif/test > @oclif/core > ts-node@10.9.2" has unmet peer dependency "@types/node@*".
warning "sfdx-git-delta > @salesforce/command > @oclif/test > @oclif/core > ts-node@10.9.2" has unmet peer dependency "typescript@>=2.7".
Installing plugin sfdx-git-delta... installed v5.42.0-dev-901.10410862637-1
arielsi@Ariel:/mnt/c/Users/arielsi/IdeaProjects/SgdFlowTranslations$ sfdx plugins
 ›   Warning: sfdx-cli update available from 7.172.0 to 7.209.6.
sfdx-git-delta 5.42.0-dev-901.10410862637-1 (dev-901)
arielsi@Ariel:/mnt/c/Users/arielsi/IdeaProjects/SgdFlowTranslations$ sfdx sgd:source:delta --to master --from changeFlow --repo . --output .
 ›   Warning: sfdx-cli update available from 7.172.0 to 7.209.6.
(node:670) Warning: Deprecated environment variable: SFDX_AUTOUPDATE_DISABLE. Please use SF_AUTOUPDATE_DISABLE instead.
(Use `node --trace-warnings ...` to show where the warning was created)
(node:670) Warning: Deprecated environment variable: SFDX_DISABLE_AUTOUPDATE. Please use SF_DISABLE_AUTOUPDATE instead.
(node:670) Warning: Deprecated config name: apiVersion. Please use org-api-version instead.
{
  "error": null,
  "output": ".",
  "success": true,
  "warnings": []
}

scolladon:
> did you pull my last changes in my repo? this is what i do, and the generated package.xml is like in the repo.

Yep @ArielS1, I created a new repository and tried with the version of this PR.
The issue you have that I don't is for you the output folder is the current folder.
The plugin check for the translations file that are not in the output folder because when using `-d` file are copied and it should not evaluate the file generated by the plugin. In your scenario the output folder is the repository folder so translation files are ignored.
You should use a dedicated folder for the output of the plugin and avoid using the same folder as the repository because using `-d` would override in file metadata (like workflow, custom labels, etc).

You can workaround this issue by creating a folder `output` and pass this folder to the parameter `--output` :
```sh 
mkdir output
sfdx sgd:source:delta --to master --from changeFlow --repo . -o output
```

I'll ship a fix in this PR to ignore the "output folder filter" when `--output` is equal to `--repo`

scolladon:
I fixed the issue when output is equal to the repo, could you QA one more time and let me know your finding please @ArielS1 ?

scolladon:
> I fixed the issue when output is equal to the repo, could you QA one more time and let me know your finding please @ArielS1 ?

Hi @ArielS1 😁 do you think you could have a final round of UAT (hopefully) please ?

ArielS1:
> > I fixed the issue when output is equal to the repo, could you QA one more time and let me know your finding please @ArielS1 ?
> 
> Hi @ArielS1 😁 do you think you could have a final round of UAT (hopefully) please ?

Hi,

I checked and it works both on my local env and on the real build env.

Thanks

PR #910, opened 2024-09-04
docs: updates the `sf` command on the `README.md`

renatoliveira:
This PR updates the commands used in the README to the newest `sf` command format.

Note: At L353 I added a bash line for the creation of an empty package.xml file because the deploy command requires a manifest.

PR #912, opened 2024-09-04
docs: updates the `sf` command on the `README.md`

scolladon:
# Explain your changes

---

Description taken from #910: 

This PR updates the commands used in the README to the newest sf command format.

Note: At L353 I added a bash line for the creation of an empty package.xml file because the deploy command requires a manifest.

---

closes #910


README.md
```diff
@@ -326,12 +324,12 @@ Note: it is also possible to generate a **source** folder containing added/chang
...
 
 ```sh
-sfdx force:source:deploy -x package/package.xml --postdestructivechanges destructiveChanges/destructiveChanges.xml
+sf project deploy start -x package/package.xml --post-destructive-changes destructiveChanges/destructiveChanges.xml
 ```
 
 And voilà! 🥳
 
-However, keep in mind that the above command will fail if the destructive change was supposed to be executed before the deployment (i.e. as `--predestructivechanges`), or if a warning occurs during deployment. Make sure to protect your CI/CD pipeline from those scenarios, so that it doesn't get stuck by a failed destructive change.
+However, keep in mind that the above command will fail if the destructive change was supposed to be executed before the deployment (i.e. as `--pre-destructive-changes=`), or if a warning occurs during deployment. Make sure to protect your CI/CD pipeline from those scenarios, so that it doesn't get stuck by a failed destructive change.
```
mehdicherf, line 332:
```suggestion
However, keep in mind that the above command will fail if the destructive change was supposed to be executed before the deployment (i.e. as `--pre-destructive-changes`), or if a warning occurs during deployment. Make sure to protect your CI/CD pipeline from those scenarios, so that it doesn't get stuck by a failed destructive change.
```

PR #915, opened 2024-09-12
feat: add FeatureParameter metadata support

aljaz823:
# Explain your changes

---

Add metadata support for all three types of Feature Parameters:
- [FeatureParameterBoolean](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_featureparameterboolean.htm)
- [FeatureParameterDate](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_featureparameterdate.htm)
- [FeatureParameterInteger](https://developer.salesforce.com/docs/atlas.en-us.api_meta.meta/api_meta/meta_featureparameterinteger.htm)

### DIFF
--- a/__tests__/integration/services.test.ts
+++ b/__tests__/integration/services.test.ts
@@ -437,6 +437,21 @@ const testContext = [
     new Set(['WaveXmdTest']),
     'WaveXmd',
   ],
+  [
+    'force-app/main/default/featureParameters/FeatureParameterIntegerTest.featureParameterInteger',
+    new Set(['FeatureParameterIntegerTest']),
+    'FeatureParameterInteger',
+  ],
+  [
+    'force-app/main/default/featureParameters/FeatureParameterBooleanTest.featureParameterBoolean',
+    new Set(['FeatureParameterBooleanTest']),
+    'FeatureParameterBoolean',
+  ],
+  [
+    'force-app/main/default/featureParameters/FeatureParameterDateTest.featureParameterDate',
+    new Set(['FeatureParameterDateTest']),
+    'FeatureParameterDate',
+  ],
 ]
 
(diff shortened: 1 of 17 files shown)


scolladon (changes requested):
Nice addition, thanks !!!

Could the definition be at the top of the file ?
Also, has it been tested manually ?

aljaz823:
> Nice addition, thanks !!!
> 
> Could the definition be at the top of the file ? Also, has it been tested manually ?

Sure, I moved definitions to the top of metadata files.
It has been tested manually, here is the result:
image

Generated diff package.xml file:
```xml
    <types>
        <members>FeatureBoolean1</members>
        <name>FeatureParameterBoolean</name>
    </types>
    <types>
        <members>FeatureDate1</members>
        <name>FeatureParameterDate</name>
    </types>
    <types>
        <members>FeatureSet1</members>
        <members>FeatureSet2</members>
        <name>FeatureParameterInteger</name>
    </types>
    <version>60.0</version>
</Package>
```

scolladon (commented):
Thanks for putting the definition on top of the file.
The bottom is reserved to certain types I prefer to be grouped.

Could you add an [integration test](https://github.com/scolladon/sfdx-git-delta/blob/main/__tests__/integration/services.test.ts) for those types please ?
You just have to add an item in the list for each new type.
If you struggle with that I can add it, just let me know
As they are regrouped in the same folder I would like to couple the test with the metadata definition, just to fix the spec explicitely

aljaz823:
I added the integration tests, though I'm not familiar with that so I just followed other patterns. Let me know if it needs additional fixing.

scolladon (approved):
LGTM

Great job, thanks for this contribution

Issue #961, opened 2024-12-10
Lack of several metadata types in package.xml


ArielS1:
Yesterday we upgraded the repo to v5.50.0 from v5.42.1 and we noticed that the package.xml that was generated on the same change was smaller, 
and several metadata type changed weren't detected even that we saw in the logs that the script detected changes in those files.

for example:
CustomLabel
CustomObjectTranslation
Profile
SharingRules
SharingCriteriaRule
StandardValueSetTranslation
Translation

then we installed the previous version v5.49.4 and it worked correctly

scolladon:
Hi @ArielS1 !

Thanks for raising this issue and thanks for contributing in making this project better!

If I understand, in the v5.50.0 (`latest`) of the plugin there is no more detection of those metadata type, where in v5.42.1 (or `stable`) it works, right ?
If so, could you help me reproduce locally with your specific use case please ?

scolladon:
> Did you try running the script on v5.50.0 on one of the metadata types i mentioned?

Yes ! We [have a lot](https://github.com/scolladon/sfdx-git-delta/actions/runs/12231337086/job/34115671738) of unit test / integration test and functional test.
That is why I ask for way to reproduce the issue as you see it (maybe with particular metadata or operating system or environment or file structure)

Could you help me reproduce the issue locally ?
You can either describe what commit content I should create inside a repository, or you could also fork the [reproduction playground](https://github.com/scolladon/sfdx-git-delta-reproduction-playground), create a branch, write the necessary commit to reproduce and share that branch here.
And give me the commands to run with sgd to reproduce the issue.

scolladon:
Hi @ArielS1 !

Kind follow up to my previous message [here](https://github.com/scolladon/sfdx-git-delta/issues/961#issuecomment-2532070067)

Do you have a way to reproduce the issue locally ?
I'd like to investigate and fix the potential issue here ?

scolladon:
Hi @ArielS1 !

One last follow up on this matter.
Do you have more information so I can reproduce the issue ? Any chance you could answer my previous messages and give me more details and detailed reproduction scenario ?
:)

thomas100z:
Hi @scolladon,

I think I am experiencing something similar.  For me profiles are not included. However only for a specific command;
-  `sfdx sgd:source:delta --to "HEAD" --from "origin/staging" --output changed-sources/ --generate-delta --source force-app/ `  Includes profiles as expected.
[...]

PR #985, opened 2025-01-11
fix: check for commit ref for `--to` and `--from` parameters

scolladon:
# Explain your changes

---

### Fixes an issue with how `rev-parse` was used to check `--to` or `--from` parameters content
Before it was possible to pass "." or ".." to either `--to` or `--from`. Our `rev-parse` check did not return errors but was not returning parsed revision as well.
This could lead to very strange behaviour and could mess with inFile comparison and "include" feature handling.
Now the plugin uses the parameter `--verify` for the `rev-parse` command which checks only for revision or alias of revision

### Improves how error are communicate to end users.
Before error where not displayed when not using `--json`
Leaving the user interpret the result based on the exit code (not the most user friendly)
Now the plugin display the error with its message in the end or the success !

### Improves CI speed
Disable SF CLI autoupdate and telemetry

### DIFF
--- a/.github/linters/.cspell.json
+++ b/.github/linters/.cspell.json
@@ -20,6 +20,7 @@
         "appli",
         "authprovider",
         "authproviders",
+        "AUTOUPDATE",
         "behaviour",
         "benchmarkjs",
         "biomejs",
--- a/.github/workflows/on-main-push.yml
+++ b/.github/workflows/on-main-push.yml
@@ -69,11 +69,16 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 18
+      
+      - uses: google/wireit@setup-github-actions-caching/v2
 
       - name: Setup dependencies, cache and install
         uses: ./.github/actions/install
 
-      - uses: google/wireit@setup-github-actions-caching/v2
+      - name: Set environment variables
+        run: |
+          echo "SF_DISABLE_AUTOUPDATE=true" >> $GITHUB_ENV
+          echo "SF_DISABLE_TELEMETRY=true" >> $GITHUB_ENV    
 
       - name: Build plugin
         run: npm pack
--- a/.github/workflows/on-main-push.yml
+++ b/.github/workflows/on-main-push.yml
@@ -69,7 +69,7 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version: 18
-      
+
       - uses: google/wireit@setup-github-actions-caching/v2
 
       - name: Setup dependencies, cache and install
@@ -77,8 +77,8 @@ jobs:
 
       - name: Set environment variables
         run: |
-          echo "SF_DISABLE_AUTOUPDATE=true" >> $GITHUB_ENV
-          echo "SF_DISABLE_TELEMETRY=true" >> $GITHUB_ENV    
+          echo "SF_DISABLE_AUTOUPDATE=true" >> "$GITHUB_ENV"
+          echo "SF_DISABLE_TELEMETRY=true" >> "$GITHUB_ENV"
 
       - name: Build plugin
         run: npm pack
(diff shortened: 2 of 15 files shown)


scolladon:
You may want to test this @nvuillam 
Maybe your users will be impacted as this PR propose to include uncommitted work by default when `--to` parameter is not specified (or is specified with "")
Your opinion will be highly valuable as well, feel free to speak your mind !

scolladon:
You can follow those [steps](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) to easily test this PR locally

nvuillam:
> You may want to test this @nvuillam 
> Maybe your users will be impacted as this PR propose to include uncommitted work by default when `--to` parameter is not specified (or is specified with "")
> Your opinion will be highly valuable as well, feel free to speak your mind !

Sfdx-hardis always uses --to argument so if you don't change its behavior when a value is sent, it's not a breaking change so ok for me 😊

nvuillam:
Additional note: if it changes the behavior when --to is not sent, if u respect semver your next release will be 7.0.0 😊

scolladon:
> Additional note: if it changes the behavior when --to is not sent, if u respect semver your next release will be 7.0.0 😊

That's right, and it would have been the case with the previous version of this PR.
I rolledback the previous default behaviour and just allow to pass "" as a parameter if one wants to include uncommited work as well.
Enough breaking change already for january ;)
Thanks @nvuillam !

scolladon:
Quick follow up here of your tests @thomas100z 

Can you tell me how to reproduce your remarks [here](https://github.com/scolladon/sfdx-git-delta/issues/961#issuecomment-2586952174) ("new file are not added")
Just to be sure to understand properly your POV

thomas100z:
What I was doing comes down to these steps:

1. Make new CustomField in sandbox
2. Retrieve CustomField metadata from sandbox
3. Generate delta package with the `dev-985` version with command ` sf sgd source delta --to ""  --from "origin/staging" --output-dir  changed-sources/ --source-dir force-app/ `
4. Observe new CustomeField not in `changed-sources/package/package.xml`

scolladon:
Unfortunately it is not possible to `git diff` with content not even staged...
To have them included you have to add those files in the git staging area (`git add ...`).

This makes me wonder if we really need this or if just using the workaround to commit before running sgd would be enough.

**With the current change on `--to`**
If you want to compare commits : use sgd as usual
If you want to compare commits and include changes not commited : use sgd with "" for `--to``
If you want to compare commits and include changes not commited and new fils : git add new files and use sgd with "" for `--to`

-> Unstage new files if not needed finally

**Without the current change on `--to`**
If you want to compare commits : use sgd as usual
If you want to compare commits and include changes not commited : git add and git commit and sgd as usual (reset commit if needed)
If you want to compare commits and include changes not commited and new fils : git add and git commit and sgd as usual (reset commit if needed)

-> Soft reset in the end, instead of unstaging new files

Your opinion about this @mehdicherf ?

mehdicherf:
My 2 cents:

I didn't even think about that use case where `--to` is not specified. 
The tool was primarily designed to help with deployments, where anyway we only deal with metadata committed to the repo.

My personal opinion is that there is little value in offering a diff to staged (because as you mentioned, there needs to be a step to stage the files, and currently the user can simply do an additional commit (and reset later) if needed.

That being said, since you already did the job, I don't have see any objections either to include this feature in the plugin, as long as it's not a breaking change for existing situations where   `--to` is specified (my understanding is that this behaviour is unchanged, and I'm good with it!)

scolladon:
Ok I will roll back the new feature to take the uncommitted work.

This PR and the related discussions were very helpful.
It helped ensure the goal of SGD is to compare code base committed in the repository (code we actually validate and want to be deployed)

It also helped find a way to deploy uncommitted changes and new files:
- commit them, use sgd, then reset (soft if you want to keep the changes) the commit 
- create a branch locally, commit them, use sgd, then delete the branch

mehdicherf (approved):

PR #1084, opened 2025-06-04
feat: support multiple --source-dir flags

viniciushiga:
# Explain your changes

---

This PR updates the `--source-dir` option to accept multiple values instead of just one.
This makes it possible to specify several source directories in a single command. It still works the same way if you only provide one directory.

I also removed the `exists: true` check from the `--source-dir` option. This validation was incorrect for two reasons:
1. It was checking paths relative to where the command was run, but `--source-dir` is actually relative to `--repo-dir`.
2. It blocked the use of deleted directories, which could be useful if you want to build destructive changes for them.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,7 +162,7 @@ FLAGS
   -n, --include-file=<value>              file listing paths to explicitly include for any diff actions
   -o, --output-dir=<value>                [default: ./output] source package specific output
   -r, --repo-dir=<value>                  [default: ./] git repository location
-  -s, --source-dir=<value>...             [default: ./] source folder focus location related to --repo-dir
+  -s, --source-dir=<value>...             [default: ./] source folders focus location relative to --repo-dir
   -t, --to=<value>                        [default: HEAD] commit sha to where the diff is done
       --ignore=<value>                    /!\ deprecated, use '--ignore-file' instead.
       --ignore-destructive=<value>        /!\ deprecated, use '--ignore-destructive-file' instead.
@@ -193,6 +193,16 @@ FLAG DESCRIPTIONS
     salesforce metadata API version, default to sfdx-project.json "sourceApiVersion" attribute or latest version
 
     Override the api version used for api requests made by this command
+
+  -s, --source-dir=<value>...  source folders focus location relative to --repo-dir
+
+    You can use this flag multiple times to include different folders that contain source files. Each path should be
+    relative to --repo-dir.
+
+    The folder can exist or not.
+    * If the folder exists, its contents will be processed.
+    * If the folder doesn't exist, it usually won't show any output—unless the folder was recently deleted and is part
+    of a diff, in which case changes may still be picked up.
 ```
 
 _See code: [src/commands/sgd/source/delta.ts](https://github.com/scolladon/sfdx-git-delta/blob/main/src/commands/sgd/source/delta.ts)_
--- a/messages/delta.md
+++ b/messages/delta.md
@@ -34,7 +34,15 @@ source package specific output
 
 # flags.source.summary
 
-source folder focus location related to --repo-dir
+source folders focus location relative to --repo-dir
+
+# flags.source.description
+
+You can use this flag multiple times to include different folders that contain source files. Each path should be relative to --repo-dir.
+
+The folder can exist or not.
+* If the folder exists, its contents will be processed.
+* If the folder doesn't exist, it usually won't show any output—unless the folder was recently deleted and is part of a diff, in which case changes may still be picked up.
 
 # flags.ignore.summary
 
--- a/src/commands/sgd/source/delta.ts
+++ b/src/commands/sgd/source/delta.ts
@@ -53,6 +53,7 @@ export default class SourceDeltaGenerate extends SfCommand<SgdResult> {
     'source-dir': Flags.directory({
       char: 's',
       summary: messages.getMessage('flags.source.summary'),
+      description: messages.getMessage('flags.source.description'),
       default: [SOURCE_DEFAULT_VALUE],
       aliases: ['source'],
       deprecateAliases: true,


src/adapter/GitAdapter.ts
```diff
@@ -112,7 +112,7 @@ export default class GitAdapter {
       .map(line => treatPathSep(line))
   }
 
-  public async getFilesPath(path: string): Promise<string[]> {
+  protected async getFilesPathCached(path: string): Promise<string[]> {
```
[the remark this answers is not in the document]
viniciushiga: **question**: I just renamed this function. Should I go ahead and refactor it, or is it okay to leave it as is?

__tests__/functional/delta.nut.ts
```diff
@@ -13,6 +15,46 @@ describe('sgd source delta NUTS', () => {
...
+  it('run `e2e` tests with multiple --source-dir flags', async () => {
+    // Act
+    const result = execCmd(
+      'sgd source delta --from "origin/e2e/base" --to "origin/e2e/head" --output e2e/expected --generate-delta --repo e2e --source-dir test/create-classes --source-dir test/update-classes --source-dir test/delete-classes --json',
+      {
+        ensureExitCode: 0,
+      }
+    ).shellOutput
+
+    // Assert
```
scolladon, line 27: **praise**: 🙌 very nice assert section, very clear, thanks

README.md
```diff
@@ -148,7 +148,7 @@ Generate incremental package manifest and source content
 ```
```
scolladon, line 148: **thought**: 💭 I wonder if we should add an example here, maybe in the advanced section with a concrete use case

src/adapter/GitAdapter.ts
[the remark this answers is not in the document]
scolladon: IMHO it is ok, this method is hard to refactor to diminish the complexity, one way would be to split further its responsibility into sub function but I prefer to have it centralize as the algorithm is not straigthforward and there are small and clever local optimizations easy to understand when they are located at the same place.

src/utils/fsHelper.ts
```diff
@@ -53,11 +53,11 @@ export const pathExists = async (path: string, config: Config) => {
...
 
 export const readDir = async (
-  path: string,
+  paths: string | string[],
   config: Config
 ): Promise<string[]> => {
   const gitAdapter = GitAdapter.getInstance(config)
-  return await gitAdapter.getFilesPath(path)
+  return await gitAdapter.getFilesPath(paths)
 }
```
scolladon, line 61: **suggestion (blocking)**: 💡 the change here allows this function to move from reading **directory** to **directories**

I think (I'm maybe wrong and I'm open to discussion) the only reason here is because of the usage in [`flowTranslationProcessor`](https://github.com/scolladon/sfdx-git-delta/blob/3a37c8c52857a350b2080929ac2327afdf9843d2/src/post-processor/flowTranslationProcessor.ts#L59)
which leads to a lot of changes downstream (`GitAdapter.ts` and its test)

So, either we change the name of this methods from `readDir` to `readDirs` (and we add a tests suite for readDir for an array of string)
Or we just change the way it is called in `FlowTranslationProcessor._buildFlowDefinitionsMap` so it iterates the source array synchronously : 
```js
let allFiles: string[] = []
for(const source of this.config.source) {
  const subFiles = await readDir(source, this.work.config)
  // Use concat instead of push(...subFiles) to avoid potential argument stack depth issue
  allFiles = allFiles.concat(subFiles) 
}
```
or 
```js
const subFiles : Array<Array<string>> = []
for(const source of this.config.source) {
  const subFiles = await readDir(source, this.work.config)
  subFiles.push(subFiles)
}
// safe but slow ?
const allFiles = subFiles.flat()
```

If we go this road, maybe there are less changes accross the tests and the codes ? 🤔 (we just need to add a test suite in `FlowTranslationProcess.test.ts`, no change in `fsHelper` neither in `GitAdapter`)
What do you think ?

scolladon (commented):
I've done a first round, great job so far !!
Thanks a lot !

Let's discuss where this should be applied (flowTranslation or gitAdapter)

scolladon:
> I also removed the exists: true check from the --source-dir option. This validation was incorrect for two reasons:

That's a very good point, thanks for that.

Git behaves nicely when we pass a folder that does not exist (it is silent), probably for the reason you mentioned (the folder may be deleted and therefore present in a previous commit).

I wonder if it could confuse users when they put a wrong `source-dir`. In this case there will be no output (which is a possible outcome) and it can lead to false positive, not easy to discover...
Maybe explaining this behavior in the parameter description could be the way to address this 🤔

README.md
viniciushiga: **question:** How about changing the `Scoping delta generation to a specific folder` section to this:

### Scoping delta generation to specific folders

The `--source-dir [-s]` parameter allows you to specify one or more folders to focus on, making any other folders ignored.
It means the delta generation will only focus on the dedicated folders.

For example, consider a repository containing many sub-folders (force-app/packaged, force-app/unpackaged, etc).
This repository contains packaged (deployed via package) and unpackaged (deployed via CLI) sources.
You only want to apply delta generation for the unpackaged sources.

```sh
$ tree
.
├── force-app
│   ├── packaged
│   │   └── classes
│   │       └── PackagedClass.cls
│   └── unpackaged
│       ├── admin
│       │   └── profile
│       │       └── Admin.profile-meta.xml
│       ├── legacy
│       │   └── classes
│       │       └── LegacyClass.cls
│       └── ui
│           └── lightningExperienceThemes
│               └── MyCustomTheme.lightningExperienceTheme-meta.xml
└── ...

# scope the delta generation only to the unpackaged folder
$ sf sgd source delta --from commit --source-dir force-app/unpackaged

# scope to multiple folders if needed
$ sf sgd source delta --from commit --source-dir force-app/unpackaged/admin --source-dir force-app/unpackaged/ui
```

> The ignored patterns specified using `--ignore-file [-i]` and `--ignore-destructive-file [-D]` still apply.
> The `--source-dir` path must be relative to the `--repo-dir` path

scolladon: Good idea, to improve this section !

We coudl simplify a bit and present only an example with mutiple folders passed to `source-dir` : 

...
### Scoping delta generation to specific folders

The `--source-dir [-s]` parameter allows you to specify one or more folders to focus on, making any other folders ignored.
It means the delta generation will only focus on the dedicated folders.

For example, consider a repository containing many sub-folders (force-app/packaged, force-app/unpackaged, etc).
This repository contains packaged (deployed via package) and unpackaged (deployed via CLI) sources.
You only want to apply delta generation for the `admin` and `ui` folder of the unpackaged sources.

```sh
$ tree
.
├── force-app
│   ├── packaged
│   │   └── classes
│   │       └── PackagedClass.cls
│   └── unpackaged
│       ├── admin
│       │   └── profile
│       │       └── Admin.profile-meta.xml
│       ├── legacy
│       │   └── classes
│       │       └── LegacyClass.cls
│       └── ui
│           └── lightningExperienceThemes
│               └── MyCustomTheme.lightningExperienceTheme-meta.xml
└── ...

# scope the delta generation only to the unpackaged folder
$ sf sgd source delta --from commit --source-dir force-app/unpackaged/admin --source-dir force-app/unpackaged/ui
```
...

src/utils/fsHelper.ts
viniciushiga: Unfortunately, `IncludeProcessor.process` also calls `GitAdapter.getFilesPath`, passing `config.source` as an argument. So, for the latter suggestion, we’d need to iterate over the source array in both `IncludeProcessor` and `FlowTranslationProcessor`.

I tried to keep the changes minimal, that’s why `GitAdapter.getFilesPath` accepts either a string or string[] 😄. This way, I didn’t need to update existing tests or code.

I’ll go ahead and rename the method to `readDirs` and add tests for it.

viniciushiga:
I've made the changes and applied the suggestions. Let me know if everything looks good or if changes are needed 🙂

mehdicherf (approved):
Nice contribution @viniciushiga 

Looks good to me @scolladon

viniciushiga:
awesome! 🎉 
thanks @scolladon @mehdicherf

PR #1111, opened 2025-08-16
fix(metadata): generate Report/ReportFolder and Dashboard/DashboardFolder correctly

i-m-anas:
Previously, the package builder placed report folders under the Report type (and dashboard folders under Dashboard). This fails when migrating folder-only changes (empty folders).

Now, package.xml generation separates items and folders:
- Reports → Report
- Report folders → ReportFolder
- Dashboards → Dashboard
- Dashboard folders → DashboardFolder

This supports all combinations (folders only, items only, or both) and remains backwards compatible.

# Explain your changes

---
The package generator was previously putting both items (reports/dashboards) and their folders
under the same type (Report or Dashboard). This is valid as long as the package contains at least one
report or dashboard.

Problem
If a package contained only empty folders (for example, migrating new folder structure without any
reports/dashboards inside), the deployment would fail. Salesforce requires folders to be represented
under their own types: ReportFolder and DashboardFolder.

Change
	•	Continue to generate items under Report / Dashboard.
	•	Generate folders under their respective folder types: ReportFolder and DashboardFolder.
	•	Works in all cases:
	•	Items + folders together
	•	Items only
	•	Folders only

---
No

closes #

- [x] NUT tests updated to cover the fix.

---

# Any other comments

---

### DIFF
--- a/__tests__/functional/delta.nut.ts
+++ b/__tests__/functional/delta.nut.ts
@@ -70,8 +70,8 @@ describe('sgd source delta NUTS', () => {
     const destructiveChangesLineCount = await getFileLineNumber(
       'e2e/expected/destructiveChanges/destructiveChanges.xml'
     )
-    expect(packageLineCount).to.equal(232)
-    expect(destructiveChangesLineCount).to.equal(137)
+    expect(packageLineCount).to.equal(235)
+    expect(destructiveChangesLineCount).to.equal(140)
     expect(result).to.include('"status": 0')
   })
 })
--- a/__tests__/unit/lib/service/inNestedFolderHandler.test.ts
+++ b/__tests__/unit/lib/service/inNestedFolderHandler.test.ts
@@ -1,0 +1,182 @@
+'use strict'
+import { describe, expect, it, jest } from '@jest/globals'
+
+import { METAFILE_SUFFIX } from '../../../../src/constant/metadataConstants'
+import { MetadataRepository } from '../../../../src/metadata/MetadataRepository'
+import InNestedFolder from '../../../../src/service/inNestedFolderHandler'
+import type { Work } from '../../../../src/types/work'
+import { copyFiles, readDirs } from '../../../../src/utils/fsHelper'
+import { getGlobalMetadata, getWork } from '../../../__utils__/globalTestHelper'
+
+jest.mock('../../../../src/utils/fsHelper')
+const mockedReadDirs = jest.mocked(readDirs)
+
+const entity = 'folder/test'
+const extension = 'report'
+const objectType = {
+  directoryName: 'reports',
+  inFolder: true,
+  metaFile: true,
+  xmlName: 'Report',
+  content: [
+    {
+      suffix: 'report',
+      xmlName: 'Report',
+    },
+    {
+      suffix: 'reportFolder',
+      xmlName: 'ReportFolder',
+    },
+  ],
+}
+
+const testContext = [
+  [
+    `A       force-app/main/default/${objectType.directoryName}/${entity}.${extension}-meta.xml`,
+    new Set([entity]),
+    'Report',
+  ],
+  [
+    `A       force-app/main/default/${objectType.directoryName}/${entity}.reportFolder-meta.xml`,
+    new Set([entity]),
+    'ReportFolder',
+  ],
+  [
+    `A       force-app/main/default/${objectType.directoryName}/folder/${entity}.reportFolder-meta.xml`,
+    new Set([`folder/${entity}`]),
+    'ReportFolder',
+  ],
+  [
+    `A       force-app/main/default/${objectType.directoryName}/folder/folder/${entity}.reportFolder-meta.xml`,
+    new Set([`folder/folder/${entity}`]),
+    'ReportFolder',
+  ],
+]
+
+let work: Work
+beforeEach(() => {
+  jest.clearAllMocks()
+  work = getWork()
+})
+
+describe('InNestedFolderHandler', () => {
+  let globalMetadata: MetadataRepository
+  beforeAll(async () => {
+    globalMetadata = await getGlobalMetadata()
+  })
+
+  describe.each(testContext)(
+    'when called with generateDelta false',
+    (
+      changePath: string | Set<string>,
+      expected: string | Set<string>,
+      expectedType: string | Set<string>
+    ) => {
+      beforeEach(() => {
+        work.config.generateDelta = false
+      })
+      it(`should not copy meta files nor copy special extension when adding ${expectedType}`, async () => {
+        // Arrange
+        const sut = new InNestedFolder(
+          changePath as string,
+          objectType,
+          work,
+          globalMetadata
+        )
+
+        // Act
+        await sut.handleAddition()
+
+        // Assert
+        expect(work.diffs.package.get(expectedType as string)).toEqual(expected)
+        expect(copyFiles).not.toHaveBeenCalled()
+      })
+    }
+  )
+
+  describe.each(testContext)(
+    'when called with generateDelta true',
+    (
+      changePath: string | Set<string>,
+      expected: string | Set<string>,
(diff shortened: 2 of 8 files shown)


scolladon:
Hi @i-m-anas !

Thanks for this very nice contribution.
I'll review it and help to make it land in the repo as it definitely solve a problem and bring value.

__tests__/unit/lib/service/inNestedFolderHandler.test.ts
```diff
@@ -0,0 +1,182 @@
+'use strict'
```
scolladon, line 1: **thought**: 💭 I wonder if we could have the same result by adding `reportFolder` and `dashboardFolder` in metadata definition files

I'll try it soon to better understand

__tests__/functional/delta.nut.ts
```diff
@@ -70,8 +70,8 @@ describe('sgd source delta NUTS', () => {
     const destructiveChangesLineCount = await getFileLineNumber(
       'e2e/expected/destructiveChanges/destructiveChanges.xml'
     )
-    expect(packageLineCount).to.equal(232)
```
scolladon, line 73: **chore**: 🧹 (to myself) need to update the e2e test to match this new behavior

src/metadata/v63.ts
```diff
@@ -989,15 +989,33 @@ export default [
     directoryName: 'reports',
     inFolder: true,
     metaFile: false,
-    suffix: 'report',
```
scolladon, line 992: **question (blocking)**: ❓this behavior has been introduced since `v63` ?

Or it should be propagated to all the metadata file as well ?

src/service/inNestedFolderHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  protected override _fillPackage(store: Manifest) {
+    const type = this.sharedFolderMetadata.get(this.ext!) as string
+    fillPackageWithParameter({
+      store,
+      type: type,
+      member: this._getElementName(),
+    })
+  }
```
scolladon, line 38: **praise**: 🙌 this is where the magic happens :)

scolladon:
@i-m-anas could you explain exactly how to reproduce the issue please ?
What to commit and what should be the content of the repo ?

src/metadata/v63.ts
i-m-anas: I have tested it with v63 and v64, not sure about older than that, but I can perform quick test if that helps.

husainx:
> @i-m-anas could you explain exactly how to reproduce the issue please ? What to commit and what should be the content of the repo ?

@scolladon you will need some nested report folders without any reports in it. Something like;

```text
└── force-app
    └── main
        └── default
            └── reports
                ├── TopFolder1
                │   ├── Subfolder1
                │   │   └── NestedFolder.reportFolder-meta.xml
                │   ├── Subfolder2.reportFolder-meta.xml
                │   └── Subfolder3.reportFolder-meta.xml
                ├── TopFolder2
                │   └── Subfolder1.reportFolder-meta.xml
                └── TopFolder3.reportFolder-meta.xml
```

The slim version should also result in the same behaviour
```text
└── force-app
    └── main
        └── default
            └── reports
                ├── Folder1.reportFolder-meta.xml
                └── Folder2.reportFolder-meta.xml
```
the key is that there are no reports in it.
When you commit these and try to generate the package.xml it will create following `package.xml`
```xml
    <types>
        <!--  Members also include Reports here! -->
        <members>TopFolder1</members>
        <members>TopFolder1/Subfolder1</members>
        <members>TopFolder1/Subfolder1/NestedFolder</members>
        <members>TopFolder1/Subfolder2</members>
        <members>TopFolder1/Subfolder3</members>
        <members>TopFolder2</members>
        <members>TopFolder2/Subfolder1</members>
        <members>TopFolder3</members>
        <name>Report</name>
    </types>
    <version>64.0</version>
</Package>
``` 
When you run the validation or deployment with it, salesforce will complain about those end nodes thinking that those are reports.
eg:
```text
An object 'TopFolder1/Subfolder1/NestedFolder' of type Report was named in package.xml, but was not found in zipped directory 
```
But if we generate the the folders and reports in there specific metadata type if will perform the operation successfully
```
<!--
    All Reports goes here!
    <types>
         <members></members>
        <name>Report</name>
    </types>
 -->
    <types>
        <members>TopFolder1</members>
        <members>TopFolder1/Subfolder1</members>
        <members>TopFolder1/Subfolder1/NestedFolder</members>
        <members>TopFolder1/Subfolder2</members>
        <members>TopFolder1/Subfolder3</members>
        <members>TopFolder2</members>
        <members>TopFolder2/Subfolder1</members>
        <members>TopFolder3</members>
        <name>ReportFolder</name>
    </types>
    <version>64.0</version>
</Package>
```

NOTE: The same is true for dashboard as well.

__tests__/unit/lib/service/inNestedFolderHandler.test.ts
i-m-anas: So, I thought about that as well and tried that first, but I didn't pursue that because
1. The `reportFolders` and `report` shares the same directory (i.e. `force-app/main/default/reports`) and same is the case for `dashboard`, so it will not adhere to the same pattern that you have but in place for shared folders which are currently using `SharedFolderHandler`. 
2. Let say if we add the `reportFolder` and `dashboardFolder` metadata, then we will have to update the `InFolderHandler` implementation to force skip the folder metadata and only pick `report` and `dashboard` xml.
3. The handler for this case seemed to me as the mix of `SharedFolderHandler` and `InFolderHandler`, thats why I extended one and borrowed needed functionality  from other.

husainx:
> First read is great ! Great job, thanks
> 
> I think it may have an impact in the e2e tests But first I'd like to see if we could make it works using simple metadata declaration as reporfFolder and dashboardFolder may work the same as standard metadata. I need to poc around locally to better understand how this metadata works and get familiarized with the issue stated here.
> 
> I may have to duplicate this PR so it passes the full pipeline

Thanks let me know, if I can help.

src/service/inNestedFolderHandler.ts
```diff
@@ -0,0 +1,39 @@
...
+import { join } from 'node:path/posix'
+import { METAFILE_SUFFIX } from '../constant/metadataConstants.js'
+import { MetadataRepository } from '../metadata/MetadataRepository.js'
+import { getSharedFolderMetadata } from '../metadata/metadataManager.js'
+import { Metadata } from '../types/metadata.js'
+import type { Manifest, Work } from '../types/work.js'
+import { fillPackageWithParameter } from '../utils/packageHelper.js'
+import InFolderHandler from './inFolderHandler.js'
+
+export default class InNestedFolderHandler extends InFolderHandler {
```
scolladon, line 12: **suggestion:** I'd like to suggest another name : `ReportingFolderHandler``

I think it is ok to be close to the real type it handles because there is not much chance this kind of behavior would be used for other metadata kind

scolladon:
@i-m-anas please take in consideration my last comment
And rebase on top of the current main (I just made a release).

Then I'll create a branch on this remote to kick the full CI process and see if it needs a fix in the E2E tests
We have never been closer 🚀

scolladon (commented):
Thanks for taking the suggestion and rebasing it.

I'll work on the E2E test and all the remaining work to merge this PR

husainx:
Thanks @scolladon 
Eagerly waiting for this to go out and update my pipeline cache. 🚋 
Ping me if anything else is needed on this.

src/metadata/v63.ts
scolladon: Could you please test this behavior in other version ?

Here is how I would try : 
I'd start with v46, if it works then it works for every version and it should be applied in every metadata file
Else I'd do a binary search on the version to find the earliest version with this behavior.

We'll merge as soon as we have this behavior applied to all the versions that support it 🚀

PR #1119, opened 2025-08-22
fix: more file included for "Resource" kind metadata

### LINKED ISSUE #1117: Including more files than should for LWC components?
Issue verification check:

## What is the problem?

I have this dir structure in my SF project:

force-app / main / default / lwc / component1 (contains lwc component)
force-app / main / default / lwc / component1Card (contains related component)
force-app / main / default / lwc / component1Artile (contains another related component)
etc

So I have a base component named component1 and other related components which have the same directory name as the other but has a different suffix. 

When I make a change to the javascript file in the component1 dir, all files from all components listed above get included in the package.

sf sgd source delta --from 'HEAD~1' --to 'HEAD' --output-dir dist -d

I would expect that the package only contains the files from component1. However all files from all components where the component dir name starts with 'component1' are included.

There are too many files in the package. I change one component and components that were not changed are included.

## Steps to reproduce

I think the issue is in this function:

https://github.com/scolladon/sfdx-git-delta/blob/7e25282414e1717231a158669b4890df3dae724e/src/service/inResourceHandler.ts#L32

I did some console logging in that function:

metadataName: force-app/main/default/lwc/component1
staticResourcePath: force-app/main/default/lwc

Here's the filtering:

https://github.com/scolladon/sfdx-git-delta/blob/7e25282414e1717231a158669b4890df3dae724e/src/service/inResourceHandler.ts#L41

Because the filter doesn't include the path separator suffix, it will match any directory that starts with the same directory as the one with the change.

## Comments

Is this a bug or expected behaviour? If it's a bug I can try to make a PR to fix it.

scolladon:
# Explain your changes

---

Improve the algorithm used to detect metadata files related to the file currently evaluated
Improve spec for maintainability

---

closes #1117 

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@oclif/core": "^4.5.2",
-        "@salesforce/core": "^8.20.0",
+        "@salesforce/core": "^8.21.1",
         "@salesforce/sf-plugins-core": "^12.2.4",
         "async": "^3.2.6",
         "fast-equals": "^5.2.2",
@@ -42,7 +42,7 @@
         "@types/mocha": "^10.0.10",
         "@types/node": "^24.3.0",
         "benchmark": "^2.1.4",
-        "chai": "^5.3.1",
+        "chai": "^6.0.0",
         "husky": "^9.1.7",
         "jest": "^30.0.5",
         "jest-util": "^30.0.5",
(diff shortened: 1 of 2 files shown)


scolladon:
Hi @matdodgson !

matdodgson:
If I modify one LWC it now only packages that LWC and no other related ones. Thanks @scolladon for the excellent work!

PR #1160, opened 2025-12-19
feat(main): flow deletion adds an entry for each flow version to dest…

mdornescu:
…ructiveChanges

# Explain your changes

---

No
---

closes #

- [ ] Jest tests added to cover the fix.

---

# Any other comments

---


scolladon:
Hi @mdornescu !

Thanks for this contribution, it is really appreciated.

Let's open the discussion here, flow deletion is a vast topic !
We worked with @amtrack in this [PR](https://github.com/scolladon/sfdx-git-delta/pull/893) and decided to just warn the users.

We do not want to do more than what is tracked by git so this is why we do not try to infer the flow version and we prefer let the version system to tell what should be deleted.
Salesforce says a flow cannot have more than 50 versions, but those version could start at 232... And this could be different from a sandbox to another...
This is why we prefer to warn the users and let them use [more appropriate tools to delete their flow](https://sfdx-hardis.cloudity.com/hardis/org/purge/flow/).

IMHO we should close this PR without merging it and keep the actual way to deal with flows

PR #1187, opened 2026-02-04
feat!: rely on `@salesforce/source-deploy-retrieve` metadata registry

scolladon:
# Explain your changes

---

## Dynamic Metadata Registry from SDR

Previously, sfdx-git-delta maintained its own metadata type definitions in version-specific files (`v57.ts`, `v58.ts`, etc.). This approach required manual updates whenever Salesforce released new API versions or metadata types.

**Now**, the plugin leverages `@salesforce/source-deploy-retrieve` (SDR) as the single source of truth for metadata definitions:

- **Automatic type support**: New metadata types added by Salesforce are automatically available without plugin updates
- **Reduced maintenance**: No need to manually maintain version-specific metadata files
- **Consistency**: Uses the same metadata definitions as Salesforce CLI and VS Code extensions

The `SDRMetadataAdapter` class converts SDR's registry format to the internal format used by sfdx-git-delta, handling edge cases like folder types, child types, and special metadata behaviors.

## Dynamic API Version Detection

The plugin now retrieves the latest supported API version directly from SDR instead of hardcoding it:

```typescript
const latestVersion = await SDRMetadataAdapter.getLatestApiVersion()
```

This ensures the plugin always operates with the most recent API version supported by your installed SDR package.

## Custom Metadata Type Support

A new `--additional-metadata-registry` parameter allows users to define custom metadata types not yet available in SDR:

```bash
sf sgd source delta --from HEAD~1 --additional-metadata-registry ./custom-metadata.json
```

The JSON file follows the internal metadata format:

```json
[
  {
    "xmlName": "CustomMetadataType",
    "directoryName": "customTypes",
    "suffix": "cmt",
    "inFolder": false,
    "metaFile": true
  }
]
```

**Priority order** for metadata definitions:
1. Internal registry (overrides for special handling)
2. SDR registry (standard Salesforce types)
3. Additional metadata registry (user-defined types)

## Automated SDR Updates via Dependabot

To ensure timely support for new Salesforce metadata types, a dedicated CI workflow automatically handles SDR updates:

**Workflow: `on-dependency-update.yml`**
- Triggers when `package.json` or `package-lock.json` changes on PRs to main
- Runs a lightweight build validation (no full test suite)
- Enables fast iteration on dependency updates

**Dependabot configuration** (`.github/dependabot.yml`):
```yaml
version: 2
updates:
  - package-ecosystem: "npm"
    directory: "/"
    schedule:
      interval: "weekly"
    allow:
      - dependency-name: "@salesforce/source-deploy-retrieve"
```

When SDR releases a new version with updated metadata types:
1. Dependabot creates a PR automatically
2. The lightweight CI validates the build
3. After merge, the release workflow creates a new plugin version

This automation ensures users get support for new Salesforce metadata types within days of SDR updates, without manual intervention.

---

### O(n²) to O(1) Lookup in RepoGitDiff

Changed `deletedRenamed` from `string[]` to `Set<string>` in `_getRenamedElements()` and `_filterInternal()`:

```typescript
// Before: O(n) lookup per line
return !deletedRenamed.includes(line) && this.metadata.has(line)

// After: O(1) lookup
return !deletedRenamed.has(line) && this.metadata.has(line)
```

### Bounded Concurrency in AsyncFilter

Replaced unbounded `Promise.all()` with `async.filterLimit` to prevent resource exhaustion on large arrays:

```typescript
import { filterLimit } from 'async'
import { getConcurrencyThreshold } from './concurrencyUtils.js'

const asyncFilter = async (list: string[], predicate: (t: string) => Promise<boolean>) => {
  return filterLimit(list, getConcurrencyThreshold(), predicate)
}
```

### Centralized Concurrency Control

Created `src/utils/concurrencyUtils.ts` to centralize concurrency limits (max 6 for CI environments):

```typescript
const MAX_CONCURRENCY = 6

export const getConcurrencyThreshold = (): number => {
  const available = availableParallelism ? availableParallelism() : Infinity
  return Math.min(available, MAX_CONCURRENCY)
}
```

Used consistently in `diffLineInterpreter.ts` and `asyncFilter.ts`.

### Cached Regex Compilation

Added lazy regex caching in `standardHandler.ctor()` to avoid recompiling on every call:

```typescript
private lineRegex: RegExp
constructor(...) {
	...
    this.lineRegex = new RegExp(...)
	...
}
```

---

### Abstract BaseProcessor

Made `BaseProcessor` an abstract class with an abstract `process()` method:

```typescript
export default abstract class BaseProcessor {
  public abstract process(): Promise<void>
}
```

This provides **compile-time enforcement** that all post-processors implement the required method, replacing the previous runtime check.

### Unified Error Handling Utilities

Created `src/utils/errorUtils.ts` with utilities for safe error handling:

```typescript
export const getErrorMessage = (error: unknown): string => {
  if (error instanceof Error) {
    return error.message
  }
  return String(error)
}

export const wrapError = (message: string, cause: unknown): Error => {
  return new Error(message, { cause })
}
```

- `getErrorMessage`: Safely extracts message from any caught value (handles non-Error throws)
- `wrapError`: Creates errors with `cause` for stack trace preservation

Used consistently in `PostProcessorManager` for improved debuggability.

### FlowTranslationProcessor Type Safety

Added proper TypeScript interfaces to replace `any` types:

```typescript
interface FlowDefinition {
  fullName?: string
  [key: string]: unknown
}

interface TranslationsContent {
  '?xml'?: { '@_version': string; '@_encoding': string }
  Translations?: {
    '@_xmlns'?: string
    flowDefinitions?: FlowDefinition | FlowDefinition[]
    [key: string]: unknown
  }
}
```

This improves IDE support and catches type errors at compile time while maintaining flexibility for dynamic XML content.

### Unified XmlContent Type

Consolidated XML content typing in `fxpHelper.ts` as the single source of truth:

```typescript
export type XmlContent = Record<string, any>
```

Other files (like `metadataDiff.ts`) now import this type instead of defining their own, ensuring consistency across the codebase.

### Zod Schema Refactoring

Consolidated type definitions by using Zod schemas as the single source of truth:

- **Before**: Separate TypeScript interfaces and Zod schemas that could drift apart
- **After**: Types derived from schemas using `z.infer<typeof Schema>`

This ensures runtime validation and compile-time types are always in sync.

### MetadataDiff Performance & Design Improvements

Refactored `MetadataDiff.ts` with focus on performance, coupling, and state management:

**Performance Optimization (O(n×m) → O(n+m))**
- Pre-computed Map lookups in `MetadataComparator` instead of `find()`/`some()` calls
- Significant performance improvement for large metadata files

**Temporal Coupling Removal**
- `compare()` now returns `toContent` and `fromContent` in its result
- `prune()` accepts these as parameters instead of relying on instance state
- Methods can now be called independently or in different orders

**Mutable State Elimination**
- Removed instance-level `isEmpty` flag and `checkEmpty()` method
- `JsonTransformer` now uses pure functions returning `PartialResult`
- Better testability and thread-safety

```typescript
// Before: Temporal coupling
await metadataDiff.compare(path)
const result = metadataDiff.prune()

// After: Explicit data flow
const { toContent, fromContent } = await metadataDiff.compare(path)
const { xmlContent, isEmpty } = metadataDiff.prune(toContent, fromContent)
```

### Lazy Logging Consistency

Ensured consistent use of the `lazy` template tag for deferred log message construction, preventing unnecessary string operations when log levels are disabled.

---

### DIFF
--- a/__tests__/__utils__/testWork.ts
+++ b/__tests__/__utils__/testWork.ts
@@ -1,16 +1,5 @@
 'use strict'
-import { MetadataRepository } from '../../src/metadata/MetadataRepository'
-import {
-  getDefinition,
-  getLatestSupportedVersion,
-} from '../../src/metadata/metadataManager'
 import type { Work } from '../../src/types/work'
-
-export const getGlobalMetadata = async (): Promise<MetadataRepository> => {
-  const apiVersion: number = await getLatestSupportedVersion()
-  const metadata: MetadataRepository = await getDefinition({ apiVersion })
-  return metadata
-}
 
 export const getWork = (): Work => ({
   diffs: {
--- a/__tests__/integration/services.test.ts
+++ b/__tests__/integration/services.test.ts
@@ -7,10 +7,10 @@ import {
   MODIFICATION,
 } from '../../src/constant/gitConstants'
 import { MetadataRepository } from '../../src/metadata/MetadataRepository'
+import { getDefinition } from '../../src/metadata/metadataManager'
 import TypeHandlerFactory from '../../src/service/typeHandlerFactory'
 import type { Work } from '../../src/types/work'
 import { pathExists, readDirs, readPathFromGit } from '../../src/utils/fsHelper'
-import { getGlobalMetadata } from '../__utils__/globalTestHelper'
 
 jest.mock('../../src/utils/fsHelper')
 
@@ -482,7 +482,7 @@ const existingFiles = [
 
 let globalMetadata: MetadataRepository
 beforeAll(async () => {
-  globalMetadata = await getGlobalMetadata()
+  globalMetadata = await getDefinition({})
 })
 let work: Work
 let handlerFactory: TypeHandlerFactory
--- a/__tests__/unit/lib/adapter/GitAdapter.test.ts
+++ b/__tests__/unit/lib/adapter/GitAdapter.test.ts
@@ -10,7 +10,7 @@ import {
   getLFSObjectContentPath,
   isLFS,
 } from '../../../../src/utils/gitLfsHelper'
-import { getWork } from '../../../__utils__/globalTestHelper'
+import { getWork } from '../../../__utils__/testWork'
 
 const mockedRaw = jest.fn()
 const mockedAddConfig = jest.fn()
--- a/__tests__/unit/lib/metadata/metadataManager.test.ts
+++ b/__tests__/unit/lib/metadata/metadataManager.test.ts
@@ -17,42 +17,12 @@ describe(`test if metadata`, () => {
   beforeEach(() => {
     resetMetadataCache()
   })
-  it('provide latest when apiVersion is undefined', async () => {
-    const metadata = await getDefinition({ apiVersion: undefined })
-    const latestVersionSupported = await getLatestSupportedVersion()
-    const latestMetadataDef = await getDefinition({
-      apiVersion: latestVersionSupported,
-    })
+  it('Given no config, When getDefinition, Then returns metadata with known types', async () => {
+    const metadata = await getDefinition({})
 
     expect(metadata).toBeDefined()
-    expect(metadata).toEqual(latestMetadataDef)
-    expect(latestMetadataDef.get('classes')).toBeDefined()
-    expect(latestMetadataDef.get('do not exist')).toBeUndefined()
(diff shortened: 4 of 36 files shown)


scolladon:
Splitted in multiple smaller PRs

PR #1206, opened 2026-02-14
ci: update e2e expected files on SDR upgrade

scolladon:
## Summary
- Adds a new step to the Dependabot auto-merge workflow that updates API version in e2e expected XML files when `@salesforce/source-deploy-retrieve` is upgraded
- Uses `git worktree` to check out the `e2e/head` branch without disturbing the PR branch
- Updates `<version>` tags in `expected/package` and `expected/destructiveChanges` XML files to match the new SDR API version
- Amends the last commit on `e2e/head` with `--force-with-lease` for safe push

## Context
When Dependabot upgrades SDR, the API version can change (e.g. 65→66). The current workflow syncs the internal registry but does NOT update the e2e expected XML files, causing e2e test failures on `<version>` mismatch.

## Test plan
- [ ] Review the workflow YAML for correct step ordering and syntax
- [ ] Verify the `if` condition matches existing SDR-specific steps
- [ ] Confirm `getCurrentApiVersion()` returns a number (append `.0` for `<version>66.0</version>` format)
- [ ] Validate worktree cleanup happens even on success path

### DIFF
--- a/.github/workflows/dependabot-auto-merge.yml
+++ b/.github/workflows/dependabot-auto-merge.yml
@@ -46,6 +46,33 @@ jobs:
           npm run build
           npm run lint:fix
 
+      - name: Update e2e expected files API version
+        if: steps.metadata.outputs.package-ecosystem == 'npm' && contains(steps.metadata.outputs.dependency-names, '@salesforce/source-deploy-retrieve')
+        run: |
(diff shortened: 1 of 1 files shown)


scolladon:
Merged into chore/codebase-audit-fixes (PR #1205) instead.

PR #1274, opened 2026-04-12
ci(perf): post same-runner perf comparison as PR comment

scolladon:
## Explain your changes

---

`compareBaseline.mjs` now posts its same-runner performance comparison as a PR comment via the GitHub API. Uses a marker comment (`<!-- same-runner-perf -->`) to update in place on re-runs.

The workflow passes `GITHUB_TOKEN` and `PR_NUMBER` env vars to the compare step. This is the authoritative perf comparison (same runner, no CI noise). The benchmark-action gh-pages data is only used for the dashboard.

---

N/A

- [ ] Vitest tests added to cover the fix.

---

N/A — CI-only change. Verify PR comment appears on next PR with perf benchmarks.

## Any other comments

---

Aligned with the same change in sf-git-merge-driver, apex-mutation-testing, and dataset-loader.


scolladon:
Replaced by new clean branch — old branch had too many conflicts from already-merged commits.

PR #1281, opened 2026-04-22
feat: --changes-manifest for review-centric change-kind output

### LINKED ISSUE #1128: Differentiate between added and modified changed metadata
Feature Request: Differentiate between added and modified changed metadata
Is your proposal related to a problem?
First, thank you for creating and maintaining this tool! It's been great, well-maintained, well-documented and performant on big repos.

My primary use case is being able to identify the delta of metadata changes over time for review purposes. Usually in the context of a user story or epic, containing many commits over time across people. Before pushing out a new build we can easily see the delta between versions. While the tool is great for this, I've found the review process is hampered by the fact that (for obvious reasons in the package.xml) newly added (A) and modified (M) metadat components are grouped together. This makes it difficult to quickly distinguish between a brand new component and one that was simply updated, which is an important distinction for our reviews.

Describe a solution you propose
I propose adding a feature that would allow the delta generation to be filtered by change type. This would make it possible to generate a package containing only newly added (A) files, only modified (M) files, or both.

This would give users more granular control over the output and better support review-centric workflows.

Describe alternatives you've considered
To address this need, I have already created a fork of the repository and implemented this logic. My version successfully allows me to pass a filter to differentiate between the change types.

My implementation can be viewed here: https://github.com/stefanvdk/sfdx-git-delta

Additional context
Before creating a PR, I wanted to open this issue to see if you would be open to accommodating this use case in the main package. And if so, if you like this solution direction. I believe this could be a valuable feature for other developers working on the same repo’s.

Thank you again for your time and for this excellent tool!

scolladon:
## Explain your changes

ISV managed-package workflows need to distinguish **newly added** components (a 1-way door for subscribers) from **modified** ones at review time. `package.xml` deliberately bundles both for deployment. This PR adds an opt-in JSON sidecar that groups components by change kind, alongside the untouched xml manifests.

### User-facing change

New flag `--changes-manifest [-c]`:
- Bare form writes to `<output-dir>/changes.manifest.json`.
- Explicit path (`-c reports/changes.json`) resolves against cwd, same convention as `--ignore-file`.
- Absent → feature off; xml output byte-identical to pre-feature runs.

When set, it also enables git's `-M` rename detection so file-level renames surface as a dedicated `rename` bucket instead of a delete+add pair.

JSON shape:

```json
{
  "add":    { "ApexClass":   ["BrandNewClass"] },
  "modify": { "CustomLabels":["MyApp.Label1"] },
  "delete": { "ApexTrigger": ["OldTrigger"] },
  "rename": { "ApexClass":   [{ "from": "OldName", "to": "NewName" }] }
}
```

### Architectural change — `ChangeSet`

Replaces the ad-hoc `diffs`/`changes` pair with a single domain object storing two orthogonal axes:
- `byTarget: Record<ManifestTarget, Manifest>` drives `package.xml` / `destructiveChanges.xml`.
- `byKind: Record<AddKind, Manifest>` drives the JSON sidecar.

Both axes are needed: `InFileHandler` stamps `(target=Package, changeKind=Delete)` when a container file is deleted but sub-elements survive — the deployment must list it in `package.xml` while reviewers still see a delete in the JSON. Pure-projection views (`forPackageManifest`, `forDestructiveManifest`, `byChangeKind`) expose each consumer's axis; deletion cancellation and rename-participant removal are applied on read.

Rename detection is **flag-gated**. Default sgd runs still pass `--no-renames` + `--diff-filter=AMD` (fast path) or three numstat calls without the R-filter pass (whitespace path); `--changes-manifest` opts into `-M` + `--diff-filter=AMDR` plus the fourth `--numstat -M -z --diff-filter=R` call.

### Cycles of review

- Cycle 1: narrow `ManifestElement.changeKind` to `AddKind`; defensive Rename skip at the factory; `Logger.warn` for `RenameResolver` skip paths; non-`Error` throw narrowing in `ConfigValidator`.
- Cycle 2: type-level exclusion of `Rename` from `ChangeSet.from` input.
- Cycle 3: doc alignment + argv-surgery unit tests for the bare-flag parser + functional test for a rename flowing through `RenameResolver`.
- Cycle 4: Stryker pass; residual survivors are equivalent mutants (lazy-log template literals, defence-in-depth `from===to` guard).
- Cycle 5: `byTarget`/`byKind` split to restore e2e parity on `InFileHandler` container-deletion semantics + `-M` gating on `config.changesManifest`.

closes #1128

- [x] Vitest tests added (100% line/branch/function coverage holds across feature files).
- [x] NUT tests added (`__tests__/integration/commands/sgd/source/delta.nut.ts`).
- [x] E2E — existing scenarios byte-identical; no new e2e baseline drift.

```sh
# Bare (writes to <output>/changes.manifest.json)
sf sgd source delta --from HEAD~5 --to HEAD --output-dir incremental --changes-manifest

# Explicit path
sf sgd source delta --from HEAD~5 --to HEAD --changes-manifest reports/changes.json

# Verify xml is unchanged when flag absent (rename detection off)
sf sgd source delta --from HEAD~5 --to HEAD --output-dir incremental
```

Compare the xml output across flag-present vs flag-absent runs — should be byte-identical. The JSON only appears when the flag is set.

## Any other comments

- `manifestAggregator` utility removed; `ChangeSet.from(handlerResult.manifests)` is the single ingestion point in `main.ts`.
- Mutation score after cycle 4 on feature files: `changesManifestProcessor` 100%, `changeSet` 98.4%, `GitAdapter` 89.5%, `renameResolver` 85% (residuals documented as equivalent mutants).
- 1251 unit tests pass on the final state.


scolladon:
Hi @Stefanvdk !

You can follow those [steps](https://github.com/scolladon/sfdx-git-delta/blob/main/CONTRIBUTING.md#testing-the-plugin-from-a-pull-request) to easily do it locally using the draft release:
```sh
sf plugins install https://github.com/scolladon/sfdx-git-delta/releases/download/dev-pr-1281/sfdx-git-delta-dev-pr-1281.tgz
```

Stefanvdk:
Hi @scolladon 

### Context
  We maintain a fork of sfdx-git-delta ([stefanvdk/sfdx-git-delta](https://github.com/Stefanvdk/sfdx-git-delta))
  that powers an internal Salesforce change-review tool. We built a comparison harness to validate
  this PR against our existing approach before merging it into our fork.

  ### What we tested
  We cherry-picked the `feat/metadata-change-kind` branch on top of our fork and ran a test matrix
  against a real Salesforce repository with ~9,400 commits and 3 years of history. Six scenarios:

  - Single-commit modification
  - Delete-only commit
  - File move / rename (class moved to different package directory)
  - camelCase refactor (12 Apex classes renamed from lowercase to UpperCamelCase)
  - Small tag-to-tag delta (~8 commits)
  - Large tag-to-tag delta (~69 commits)

  For each scenario we compared our current approach (3 separate SGD calls with a `changeTypesToHandle`
  filter) against a single call with `--changes-manifest`.

  ### What we found

  **1. Parity holds across all scenarios.**
  The single-call approach produces byte-identical results to our current 3-call approach for every
  normal scenario. Safe to adopt.

  **2. Sub-element classification is more accurate.**
  In our large tag delta, two `CustomLabel` entries that were newly added inside a modified container
  file (`CustomLabels.labels-meta.xml`) were previously classified as `modify` by our approach.
  The new `byChangeKind()` correctly classifies them as `add`. Small improvement but the right call.

  **3. Rename detection works exactly as advertised.**
  - A class moved between packages: correctly surfaced as 1 rename pair instead of a phantom add + delete.
  - 12 Apex classes renamed from lowercase to UpperCamelCase: all detected and collapsed out of the
    add/delete buckets into renames.

  **4. XML output is byte-identical without the flag — confirmed.**
  We verified this against all six scenarios. Existing consumers of `package.xml` and
  `destructiveChanges.xml` are unaffected unless they opt in.

  This is a clean improvement. 
  Thanks for the thorough PR and the work you are doing on this! 🙏

scolladon:
That's awesome !

Thanks for your very quick and thorough QA on this one !

Did you notice performance improvements as well ? Compared to `latest` tag (`v6.31.0`) ?

Stefanvdk:
Happy to help!
I didn't notice a degradation of performance but tbh this wasn't something I paid attention to during testing, so I can't help you conclusively on this front.

PR #1328, opened 2026-06-12
ci: replace inline mutation script with node-based incremental runner

mcarvin8:
## Summary

- Adds `tooling/incremental-mutation.mjs` — a Node.js script that runs Stryker only against `src/**/*.ts` files changed vs the merge base, then generates `reports/mutation/comment.md` with a rich Markdown summary (score, per-file breakdown, survived mutant diffs)
- Replaces the bash one-liner in the `test:mutation:incremental` wireit command with `node tooling/incremental-mutation.mjs`, making it cross-platform and giving it access to the comment generation logic
- Updates `on-pull-request.yml` to post the generated comment to the PR as a sticky comment via `marocchino/sticky-pull-request-comment@v2` (updates in-place on reruns rather than spamming new comments)

The script is a direct port of [`incremental-mutation.mjs`](https://github.com/mcarvin8/apex-code-coverage-transformer/blob/main/scripts/incremental-mutation.mjs) from my `apex-code-coverage-transformer` repo where it has been validated in production CI.

## Notes

The husky `pre-push` hook was temporarily cleared locally (still remains intact on the remote) to push this branch — the project has pre-existing lint issues in code not touched by this PR that caused the hook to fail.

## Known limitation — fork PRs

The `marocchino/sticky-pull-request-comment` step will fail with `Resource not accessible by integration` when CI runs against a PR opened from a fork. This is a GitHub security restriction: `pull_request` events from forks receive a read-only `GITHUB_TOKEN` regardless of the `permissions` declared in the workflow, so writing PR comments is blocked from a fork.

The step to publish the comment to the PR will skip if the PR is open from a fork repo, but should run when open from the main repo. To validate this completely, the easiest path is to push this branch directly to the upstream repo and open a same-repo PR from it.

Feel free to close this PR if you choose to re-create it in your main repo.

## Test plan

- [ ] Open a PR with at least one modified `src/**/*.ts` file and confirm Stryker runs only against changed files
- [ ] Confirm `reports/mutation/comment.md` is generated and posted as a PR comment when running in this main repo
- [ ] Confirm re-running the job updates the existing comment rather than creating a new one
- [ ] Open a PR with no `src/**/*.ts` changes and confirm mutation testing is skipped with a comment noting no files changed

### DIFF
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -120,6 +120,13 @@ jobs:
       - name: Incremental Mutation testing
         run: npm run test:mutation:incremental
 
+      - name: Post mutation results comment
+        if: always() && github.event_name == 'pull_request'
+        uses: marocchino/sticky-pull-request-comment@v2
+        with:
+          header: stryker-mutation
+          path: reports/mutation/comment.md
+
       - uses: actions/upload-artifact@v7
         with:
           name: mutation-testing-reports
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -121,7 +121,7 @@ jobs:
         run: npm run test:mutation:incremental
 
       - name: Post mutation results comment
-        if: always() && github.event_name == 'pull_request'
+        if: always() && github.event_name == 'pull_request' && github.repository_owner == 'scolladon'
         uses: marocchino/sticky-pull-request-comment@v2
         with:
           header: stryker-mutation
--- a/.github/workflows/on-pull-request.yml
+++ b/.github/workflows/on-pull-request.yml
@@ -121,7 +121,7 @@ jobs:
         run: npm run test:mutation:incremental
 
       - name: Post mutation results comment
-        if: always() && github.event_name == 'pull_request' && github.repository_owner == 'scolladon'
+        if: always() && github.event_name == 'pull_request' && !github.event.pull_request.head.repo.fork
         uses: marocchino/sticky-pull-request-comment@v2
         with:
           header: stryker-mutation
--- a/package.json
+++ b/package.json
@@ -314,7 +314,7 @@
       ]
     },
     "test:mutation:incremental": {
-      "command": "FILES=$(git --no-pager diff --name-only --diff-filter=AM --merge-base origin/main src | grep '.ts$' | paste -sd ',' -) && [ -n \"$FILES\" ] && stryker run --mutate \"$FILES\" || echo 'No source files changed — skipping mutation testing'",
+      "command": "node tooling/incremental-mutation.mjs",
       "files": [
         "src/**/*.ts",
         "__tests__/**/*.ts",
@@ -322,7 +322,8 @@
         "**/tsconfig.json"
       ],
       "output": [
-        ".stryker-tmp/**"
+        ".stryker-tmp/**",
+        "reports/mutation/**"
       ],
       "dependencies": [
         "lint"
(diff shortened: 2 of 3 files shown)


mcarvin8:
@scolladon - Totally up to you if you want to include this or not — I wanted easy transparency on mutation testing results directly in PRs for my own sf plugins and thought it was worth contributing back since this repo already had the incremental mutation testing foundation in place. I know you have more fork contributors, but I think that's a limit for other actions besides this one.

You can check out this comment here for a example https://github.com/wisefoxme/apex-test-list/pull/345#issuecomment-4693345502

scolladon:
Hi @mcarvin8 !

Thanks for all the work and for bringing this.

I'll wait a bit before merging this because I'm not yet convinced.
I'm actually fine today with how it works :D

mcarvin8:
> Hi @mcarvin8 !
> 
> Thanks for all the work and for bringing this.
> 
> I'll wait a bit before merging this because I'm not yet convinced. I'm actually fine today with how it works :D

Gotcha, makes sense. If you think of a better way to run incremental mutation testing, let me know then please!

scolladon:
Hi @mcarvin8 !

I've slept a bit on it.
What I like is the idea that we can have the mutation result in the PR (see the result), not just a green or red tag.
It would allow us to compare with local execution and or act directly without having to rerun it locally (mutation testing is very slow).

As the run is executed inside the ubuntu runner, there is no need to create a portable script for it.
But locally it should be compatible with windows (which is not the case for now...). Maybe we can use `shx` or other, just to see if it is possible without having to create a javascript file (one more thing to maintain and understand)

The other aspect is how to persist the result in the PR.
For now I think I would be happy by just uploading an artifact containing the html result. It is easy enough to have a look at it and simple enough to maintain as well.
It is a less user friendly intermediate between the existing and what you propose. 

### TL;DR;
My suggestion here:
- Keep the `mutation:incremental` script inline in `package.json` but make it portable
- Upload html report result as a github action artifact instead of using PR comments

What do you think ?

mcarvin8:
> Hi @mcarvin8 !
> 
> I've slept a bit on it. What I like is the idea that we can have the mutation result in the PR (see the result), not just a green or red tag. It would allow us to compare with local execution and or act directly without having to rerun it locally (mutation testing is very slow).
> 
> As the run is executed inside the ubuntu runner, there is no need to create a portable script for it. But locally it should be compatible with windows (which is not the case for now...). Maybe we can use `shx` or other, just to see if it is possible without having to create a javascript file (one more thing to maintain and understand)
> 
> The other aspect is how to persist the result in the PR. For now I think I would be happy by just uploading an artifact containing the html result. It is easy enough to have a look at it and simple enough to maintain as well. It is a less user friendly intermediate between the existing and what you propose.
> 
> ### TL;DR;
> My suggestion here:
> 
> * Keep the `mutation:incremental` script inline in `package.json` but make it portable
> * Upload html report result as a github action artifact instead of using PR comments
> 
> What do you think ?

I'm okay with keeping the script inline in `package.json` but making it portable.

The current incremental mutation output is already being uploaded as an artifact, so if you're good with that over PR comments, you probably don't need to make any repo changes right now. I'll close this PR for now.

```yml
      - name: Incremental Mutation testing
        run: npm run test:mutation:incremental

      - uses: actions/upload-artifact@v7
        with:
          name: mutation-testing-reports
          path: reports/mutation
```

scolladon:
If you want to use this PR to make the script portable be my guest @mcarvin8 !

PR #1332, opened 2026-06-16
fix: escape XML entities in generated manifest member names

### LINKED ISSUE #1331: [BUG NAME]Invalid manifest file: manifest/package/package.xml
with version v6.31.0, code is working fine but with vs v6.44.6 same code is failing with getting below error
Error (InvalidManifest): Invalid manifest file: manifest/package/package.xml.  InvalidChar: char '&' is not expected.

scolladon:
## Explain your changes

---

The streaming XML writer introduced in #1284 interpolates element text raw, dropping the entity escaping that `xmlbuilder2` performed pre-6.31.0. Metadata member names containing `&`, `<`, or `>` produced invalid `package.xml` / `destructiveChanges.xml`, which SDR rejects:

> Error (InvalidManifest): Invalid manifest file: manifest/package/package.xml. InvalidChar: char '&' is not expected.

Fix: a small `escapeXmlText` helper (`&`→`&amp;`, `<`→`&lt;`, `>`→`&gt;`), applied via an **opt-in** `WriteOptions.escape` flag that only `PackageBuilder` sets — package.xml / destructiveChanges.xml member names are raw domain data (component names from the diff).

The shared writer stays **byte-passthrough by default**, on purpose: the Profile/PermissionSet and flow-translation callers feed it content that came from the `txml` passthrough reader (already XML-escaped). Re-escaping there would corrupt the byte-identical round-trip — a behaviour locked by the `profile-with-entities` byte-equality fixture. The read path is left untouched (no entity decode), so there is no double-unescape on round-trip.

---

closes #1331

- [ ] NUT tests added to cover the fix — N/A: manifest generation has no org-dependent behaviour here; covered by unit + e2e.
- [ ] E2E tests added to cover the fix — existing e2e suite passes byte-identical; `e2e/base` / `e2e/head` are immutable, so no special-char fixture was added.

---

```
npx vitest run __tests__/unit/lib/utils/packageHelper.test.ts -t "XML-special"
```

Member names containing `& < > " '` now render valid XML (`& < >` escaped; quotes left untouched as they are valid in element text).

## Any other comments

---

- Target release: **6.44.7** (patch, latest-rc) — unblocks the canary re-run.
- Mutation testing: 98.20% / 0 survived on the touched files. `xmlHelper.ts` is no longer excluded from Stryker (it now carries a behavioural function).
- `DESIGN.md` updated to document the opt-in escape + why the writer stays passthrough.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -498,6 +498,8 @@ The strict failure contract that `metadataDiff` depends on (txml is otherwise to
 
 XML writing is in-house (`xmlWriter.ts`). Iterative depth-first traversal with an explicit LIFO frame stack — safe under unexpectedly-deep input, cancellation-friendly. Frame chunks are batched in a `ChunkBuffer` (8 KB threshold) before flushing to the underlying `Writable` so a 3 000-element Profile prune flushes in tens of stream writes instead of thousands. Backpressure is honoured at flush time via `once(out, 'drain')`. Indent strings are cached in a lazily-extended array keyed by depth, replacing the per-frame `INDENT.repeat(depth)` allocation.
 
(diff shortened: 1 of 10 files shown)


scolladon:
Hi @pankajAcc !

Issue #1322, opened 2026-06-04
Static resource change in a file in a subdirectory is not detected


ArielS1:
## What is the problem?

I have an existing static resource with mime type application/x-zip-compressed
it has directories with subdirectories and files.
when i try to change an existing file or adding a new file there the package.xml won't detect that the static resource was changed

### What is the parameter and the value you used with it?

sf sgd source delta --to HEAD --from 'HEAD^' --repo . --output . --api-version=66.0

### What is the expected result?
```
<Package xmlns="http://soap.sforce.com/2006/04/metadata">
    <types>
        <members>alp360_ux_src</members>
        <name>StaticResource</name>
    </types>
    <version>66.0</version>
</Package>
```
### What is the actual result?
```
<Package xmlns="http://soap.sforce.com/2006/04/metadata">
    <version>66.0</version>
</Package>
```
## Steps to reproduce

i added a new example file to an existing static resource dir, i can even see it in the git output log:
create mode 100644 force-app/main/default/staticresources/alp360_ux_src/assets/icons/new.svg

## Execution context

it was reproduced in sfdx-git-delta both 6.7.0 and 6.44.3
i remember it used to work in the past in previous versions

node 20
Salesforce cli 2.136.8

scolladon:
Hi @ArielS1!

Thanks for raising this issue and thanks for contributing in making this project better!

I'm reproducing the issue, I'll ship a fix soon

PR #1229, opened 2026-02-26
fix: use deepest directory match in metadata lookup

scolladon:
# Explain your changes

`MetadataRepositoryImpl.searchByDirectory()` stopped at the first non-sub-type directory match when scanning left-to-right. When the source path contained a project folder whose name coincidentally matched a metadata type's `directoryName` (e.g. `portals` → `Portal`), the method returned that false-positive match and never reached the real metadata directory deeper in the path (e.g. `experiences` → `ExperienceBundle`).

The fix replaces the early-break strategy with a **deepest-match** strategy: iterate left-to-right and always pick the last (deepest) matching directory, stopping early only for `inFolder` types whose subfolders are user-defined names that could collide with metadata directory names.

This also removes the now-unnecessary `TYPES_WITH_SUB_TYPES` set and the orphaned `TERRITORY_MODEL_TYPE` constant.

# Does this close any currently open issues?

closes #1225

- [x] Jest tests added to cover the fix.
- [x] NUT tests added to cover the fix.
- [ ] E2E tests added to cover the fix.

# Any particular element that can be tested locally

```bash
sf sgd source delta --from <commit> --to <commit> --output-dir out \
  --source portals/experiences
```

Where the source tree has `portals/experiences/<BundleName>/...` — ExperienceBundle metadata should now be correctly identified.

# Any other comments

The `inFolder` early-stop preserves correct behavior for types like `Document` where subfolders can have arbitrary names (e.g. `documents/classes/logo.png` still resolves to `Document`, not `ApexClass`).

PR #1335, opened 2026-06-17
fix: resolve nested Wave (and virtual content-container) files dropped from package.xml

### LINKED ISSUE #1333: WaveDashboard components in nested folders dropped from package.xml between v6.31.0 and v6.44.6
After our CI pipeline began installing `sfdx-git-delta` as `@latest` (which resolved to **v6.44.6**), the generated delta `package.xml` stopped including the majority of our **WaveDashboard** components. The exact same source tree, same git range, and same command produced a correct manifest under **v6.31.0** but an incomplete one under **v6.44.6**. This caused our Salesforce validation deployment to fail because child metadata (WaveXmd) and a referencing FlexiPage were included while their parent WaveDashboard assets were missing.

### Environment
- **Plugin:** `sfdx-git-delta`
  - Working version: **6.31.0**
  - Failing version: **6.44.6** (installed via `@latest`)
- **Salesforce CLI:** `@salesforce/cli@latest`
- **API version:** `66.0`
- **Project type:** SFDX source format, single package directory force-app
- **CI:** Jenkins (Linux container), git diff-based delta generation

### Command used
```bash
sf sgd source delta \
  --from <mergeBase> \
  --to HEAD \
  -o metadata \
  --generate-delta \
  --ignore .sgdignore \
  -a 66.0 \
  -s force-app/
```

### Our current folder structure (the key detail)
Our CRM Analytics (Wave) metadata is **not** in the default flat `wave/` layout. It is organized into **type-specific sub-folders** under a nested path:

```
force-app/main/default/src-base/crma/wave/
├── app/            *.wapp-meta.xml
├── component/      *.wcomp, *.wcomp-meta.xml
├── dashboards/     *.wdash, *.wdash-meta.xml      <-- WaveDashboard source
└── xmds/           *.xmd-meta.xml                 <-- WaveXmd source
```

Example (one dashboard):
```
force-app/main/default/src-base/crma/wave/dashboards/Account_KPI_Dashboard.wdash
force-app/main/default/src-base/crma/wave/dashboards/Account_KPI_Dashboard.wdash-meta.xml
force-app/main/default/src-base/crma/wave/xmds/Account_KPI_Dashboard.xmd-meta.xml
```

For comparison, we have a small number of dashboards that **do** sit in the standard flat layout, and those are handled correctly by **both** versions:
```
force-app/main/default/wave/Seller_Homepage.wdash
force-app/main/default/wave/Seller_Homepage.wdash-meta.xml
force-app/main/default/wave/Seller_Homepage.xmd-meta.xml
```

### What we observed
| | Build A (good) | Build B (bad) |
|---|---|---|
| SGD version | 6.31.0 | 6.44.6 |
| Source content | identical | identical |
| Git range | identical | identical |
| `WaveDashboard` members in `package.xml` | **10** | **1** |
| Only surviving dashboard | — | `Seller_Homepage` (the one in the **flat** `wave/` path) |
| Deploy result | expected | **failed** |

- Under **6.44.6**, every WaveDashboard located in the nested `…/crma/wave/dashboards/` path was **omitted** from `package.xml`.
- The **only** WaveDashboard that survived (`Seller_Homepage`) is the one that also exists in the standard flat `…/default/wave/` directory.
- The corresponding **WaveXmd** entries (`…/crma/wave/xmds/*.xmd-meta.xml`) and a **FlexiPage** that references a dashboard by devName were still emitted, producing an inconsistent manifest (children/references present, parent assets absent).

### Resulting deployment errors (downstream symptom)
- Multiple `Asset doesn't exist in the org` errors for `WaveXmd` components.
- `field <…> in xmd does not have corresponding step in its parent Asset`.
- FlexiPage error: `devName [Account_KPI_Dashboard] is not a valid dashboard devName`.

All of these trace back to the parent `WaveDashboard` components being dropped from the manifest.

### Our analysis
- Source content, git range, CLI, API version, and the SGD command were **held constant**; we verified the `.wdash` content is byte-for-byte identical between the two builds (`git diff` returns empty).
- The **only** changed variable was the SGD plugin version (6.31.0 → 6.44.6).
- The behavioral difference correlates precisely with **folder layout**: dashboards in a **nested, type-subfoldered** path (`…/crma/wave/dashboards/…`) are recognized by 6.31.0 but **not** by 6.44.6, whereas dashboards in the **flat** `…/wave/…` path are recognized by both.
- This strongly suggests a change in how SGD maps file paths to the `WaveDashboard` metadata type — specifically that newer versions expect Wave files **directly** under a `wave/` directory and no longer match `.wdash` files nested inside a `wave/<subtype>/` structure (or inside a `wave/` that is itself deeply nested under custom folders).

### Questions / requests for the SGD team
1. **Was this an intentional change** in Wave/CRMA path resolution between 6.31.0 and 6.44.6? If so, which release introduced it, and is it documented in the changelog/migration notes?
2. **What is the officially supported folder layout** for Wave metadata (`WaveDashboard`, `WaveXmd`, `WaveComponent`, `WaveApplication`) for SGD to detect them reliably? Is the `dashboards/` + `xmds/` sub-folder split supported, or is the flat `wave/` layout now required?
3. Is there a **configuration option** (e.g., custom metadata path mapping, or an extension to .forceignore/SGD config) to make SGD recognize our nested layout without restructuring the repo?
4. If restructuring is required, can you confirm the **exact target layout** SGD expects so we can migrate confidently (we plan to consolidate to flat `wave/` to match the `Seller_Homepage` example that already works)?
5. Is there a **known issue / GitHub issue** we should subscribe to or reference?

### Our interim mitigation
We have **pinned** the plugin in CI to the last known-good version to unblock our release:
```bash
sf plugins install sfdx-git-delta@6.31.0
```
We would prefer to move back to a current/supported version once we understand the correct layout, so we are not locked to an old release indefinitely.

scolladon:
## Explain your changes

---

The metadata directory-walk in `MetadataRepositoryImpl.searchByDirectory` resolves a diff path to the **deepest** matching `directoryName`, stopping only at `inFolder` types and SDR content-container adapters (`bundle`/`digitalExperience`/`mixedContent`). Internal **virtual content-containers** — one directory holding several suffix-keyed sub-types (`VirtualWave`, `VirtualBot`, `VirtualDiscovery`, `VirtualModeration`; `metaFile: true` + a non-empty `content[]`) — were not recognised, so the walk descended past `wave/` into a colliding child folder name. A `.wdash` under `…/wave/dashboards/` resolved to the report **`Dashboard`** type, failed the `InFolderHandler` suffix gate (`'dashboard' !== 'wdash'`), and was silently dropped from `package.xml`.

Fix: `ownsNestedPaths` gains a structural arm — `isVirtualContentContainer(m) = m.metaFile === true && (m.content?.length ?? 0) > 0` — so the walk stops at any virtual content-container and returns it (routed to `SharedFolderHandler`, which resolves the concrete type by suffix). It keys off the container **shape**, not an xmlName list, so all four virtual containers are covered and any future one inherits the behaviour. Decomposed `CustomObject` (empty `content[]`) is unaffected and still descends into `fields/`, `listViews/`, etc.

Beyond the reported `wave/dashboards/*.wdash`, this also fixes a second live collision (`wave/components/*.wcomp` → `ApexComponent`) and a latent one (`bots/<Bot>/rules/*` → `Territory2Rule`).

Regression bisected to #1229 (v6.32.4), which changed the walk from "first non-subtype match" to "deepest match"; #1322 (v6.44.5) generalised the stop-condition to content-container adapters but did not cover virtual containers. Last good release: v6.31.0.

---

closes #1333

---

Run `sf sgd source delta` on a repo whose CRM Analytics metadata uses a type-subfoldered layout, e.g.:

```
force-app/main/default/src-base/crma/wave/dashboards/Account_KPI_Dashboard.wdash
force-app/main/default/src-base/crma/wave/components/AccountKpi.wcomp
```

The `WaveDashboard` / `WaveComponent` members now appear in the generated `package.xml` (previously dropped). Covered by:

- unit resolution tests in `__tests__/unit/lib/metadata/MetadataRepositoryImpl.test.ts` (nested collision, sibling Bot symmetry, both predicate conjuncts, flat regression guard);
- integration member-emission rows in `__tests__/integration/services.test.ts` (real registry + `TypeHandlerFactory`).

## Any other comments

---

- NUT/E2E not added: member emission is fully proven at the unit + integration layers; NUT needs a live org and the e2e non-regression fixtures are untouched.
- Mutation score 100% on the touched file (scoped Stryker run).
- `DESIGN.md` "Multi-Index Lookup" updated to document the new stop condition.
- Pre-existing, out of scope: `__tests__/integration/services.test.ts` builds its `work` object with a `diffs` field that no longer matches the `Work` type (`changes`); present on `main`, tolerated by the test runner.

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -67,7 +67,7 @@ flowchart LR
 | Index | Key | Use case |
 | ----- | --- | -------- |
 | `extIndex` | File extension (`.cls`, `.trigger`) | Primary lookup for most types |
-| `dirIndex` | Directory name (`classes`, `triggers`) | Fallback when extension is ambiguous — picks deepest match, stops at `inFolder` types and content containers (SDR adapters `bundle`/`digitalExperience`/`mixedContent`) so a nested content folder colliding with a metadata directory name (e.g. an `icons/` folder inside a StaticResource) cannot override the container type |
+| `dirIndex` | Directory name (`classes`, `triggers`) | Fallback when extension is ambiguous — picks deepest match, stops at `inFolder` types, SDR content-container adapters (`bundle`/`digitalExperience`/`mixedContent`), and virtual content-containers (internal-registry types with `metaFile: true` and a non-empty `content[]`: Wave/Bot/Discovery/Moderation) so a nested content folder colliding with a metadata directory name (e.g. an `icons/` folder inside a StaticResource, or a `dashboards/` folder inside `wave/`) cannot override the container type |
 | `xmlNameIndex` | XML name (`ApexClass`, `ApexTrigger`) | Direct lookup by type name, exposed as `getByXmlName(xmlName)` on the interface for handler-resolution and parent-type lookup |
 
 ### Key Metadata Fields
--- a/__tests__/integration/services.test.ts
+++ b/__tests__/integration/services.test.ts
@@ -522,6 +522,16 @@ const testContext = [
     'WaveXmd',
   ],
   [
+    'force-app/main/default/src-base/crma/wave/dashboards/Account_KPI_Dashboard.wdash',
+    new Set(['Account_KPI_Dashboard']),
+    'WaveDashboard',
+  ],
+  [
+    'force-app/main/default/src-base/crma/wave/components/AccountKpi.wcomp',
+    new Set(['AccountKpi']),
+    'WaveComponent',
+  ],
+  [
     'force-app/main/default/featureParameters/FeatureParameterIntegerTest.featureParameterInteger',
     new Set(['FeatureParameterIntegerTest']),
     'FeatureParameterInteger',
--- a/__tests__/unit/lib/metadata/MetadataRepositoryImpl.test.ts
+++ b/__tests__/unit/lib/metadata/MetadataRepositoryImpl.test.ts
@@ -197,7 +197,7 @@ describe('MetadataRepositoryImpl', () => {
       },
       // VirtualBot is a sibling virtual content-container; it locks the fix's
       // symmetry across all virtual content-container types — the predicate
-      // keys off shape (metaFile + non-empty content), never xmlName.
+      // keys off the non-empty content[], never xmlName.
       {
         adapter: 'matchingContentFile',
         directoryName: 'bots',
@@ -397,30 +397,6 @@ describe('MetadataRepositoryImpl', () => {
         // Assert
         expect(result).toStrictEqual(
           expect.objectContaining({ directoryName: 'wave' })
-        )
-      })
-
-      it('does not stop at a content-bearing folder whose metaFile is false', () => {
-        // Act
-        const result = sut.get(
-          'Z force-app/main/default/moderation/dashboards/site.block.rule-meta.xml'
-        )
-
-        // Assert
-        expect(result).toStrictEqual(
-          expect.objectContaining({ directoryName: 'dashboards' })
-        )
-      })
-
-      it('does not stop at a metaFile container whose content is empty', () => {
-        // Act
-        const result = sut.get(
-          'Z force-app/main/default/siteDotComSites/dashboards/aSitedotcom.site'
-        )
-
-        // Assert
-        expect(result).toStrictEqual(
-          expect.objectContaining({ directoryName: 'dashboards' })
         )
       })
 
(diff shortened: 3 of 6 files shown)


scolladon:
Hi @rohikasturi !

PR #1342, opened 2026-06-25
feat: add --scope-to-package-directories flag to limit diff to sfdx-project.json package dirs

### LINKED ISSUE #1340: [BUG] Files not in sfdx packages should be ignored
We need to commit sf-skills into the repository, and they contain metadata templates, examples... in .claude sub folders

sfdx-git-delta detects them as added items, whereas they are not located in package folders described in sfdx-project.json

I think that any file in a folder not declared as a package should be ignored :)

mcarvin8:
## Explain your changes

---

Adds a `--scope-to-package-directories` boolean flag that, when set, reads `packageDirectories` from `sfdx-project.json` and uses them as the source paths for the git diff instead of the default `./`. Gated behind a flag so existing behavior is unchanged.

When the flag is set, `_getScopeFromPackageDirectories` resolves the project file, maps each `packageDirectories[*].path` to a source path, and replaces `config.source`. If the project file is missing or `packageDirectories` is empty, source is left unchanged and the failure is debug-logged.

If you choose to accept this solution, I will additionally suggest a follow-on update in the future to include new `--ignore-package-directories` flags to allow users to further scope the diff down to specific package directories over all.

---

closes #1340

---

Pass `--scope-to-package-directories` to any `sgd source delta` invocation in a repo that has files committed outside of `packageDirectories` (e.g. `.claude/`, CI config, scripts). Those files should no longer appear in the generated package output.

## Any other comments

---

I had this idea a while back but held off until a concrete need surfaced. Issue #1340 is that case — files outside declared package directories being incorrectly flagged as added metadata.

Chose to gate behind a flag rather than change the default to avoid a breaking change for projects that intentionally process files outside `packageDirectories`.

### DIFF
--- a/__tests__/unit/lib/utils/configValidator.test.ts
+++ b/__tests__/unit/lib/utils/configValidator.test.ts
@@ -1033,6 +1033,109 @@ describe('Given a ConfigValidator', () => {
     })
   })
 
+  describe('_getScopeFromPackageDirectories', () => {
+    it('When scopeToPackageDirectories is false, Then source is unchanged and SfProject is not resolved', async () => {
+      // Arrange
+      work.config.scopeToPackageDirectories = false
+      work.config.source = ['./']
+      const sut = new ConfigValidator(work)
+
+      // Act
+      await sut['_getScopeFromPackageDirectories']()
+
+      // Assert
+      expect(work.config.source).toEqual(['./'])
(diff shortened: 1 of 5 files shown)


mcarvin8:
Closing based on https://github.com/scolladon/sfdx-git-delta/issues/1340#issuecomment-4799785670.

PR #1343, opened 2026-06-25
fix: copy objectTranslation parent file when only field translation changes

### LINKED ISSUE #1341: Cannot deploy FieldTranslations when ObjectTranslation only contains <CustomObjectTranslation tag
Issue verification check:

## What is the problem?
In our repo we use SGD to generate delta, but deploy using SF CLI - for field translation even though the objecttranslation file exist - they are excluded and this causes our deploy to fail

sf sgd source delta --to "HEAD" --from "baseline" --output-dir ./delta --source-dir force-app --generate-delta

objectranslation file should exist

failure as objectranslation file is missing

## Steps to reproduce

---

sf sgd source delta --to "HEAD" --from "baseline" --output-dir ./delta --source-dir force-app --generate-delta

sf project deploy start --dry-run -o sfdcOrg -x ./package/package.xml
            --post-destructive-changes ./destructiveChanges/destructiveChanges.xml
            --ignore-warnings --ignore-conflicts --verbose --test-level RunLocalTests --wait 300

## Execution context

---

**Operating System:** linux
github action runner

---

scolladon:
## Explain your changes

---

This PR fixes two related bugs in `ObjectTranslationHandler`.

**1. (#1341) Missing parent `objectTranslation` file for field-only changes.** Regression from the streaming I/O migration (#1284, 6.43.0): when only a child `*.fieldTranslation-meta.xml` changes, SGD lists the parent `CustomObjectTranslation` in `package.xml` but no longer copies the parent `*.objectTranslation-meta.xml` onto disk, so `sf project deploy -x package.xml` fails on the missing decomposed parent.

Root cause: `collectAddition` only emitted the parent copy when `StreamingDiff.buildWriter()` returned a writer; a field-only change leaves the parent empty-rooted → `undefined` writer → parent silently skipped. Fix: when no streaming writer exists, fall back to a raw `GitCopy` of the parent at `--to` via the inherited `_collectCopy` helper. The pruned `StreamedContent` path is kept when the parent has its own changes (no double-emit). The shared empty-skip gate in `streamingDiff.ts` is untouched. Restores 6.42.0 behaviour.

**2. Bogus `undefined.objectTranslation-meta.xml` copy for field translations.** Surfaced while reviewing #1. For every `fieldTranslation` child addition the handler also emitted a junk copy for a file literally named `undefined.objectTranslation-meta.xml`: `collectAddition` calls `StandardHandler.prototype.collectAddition` directly (to skip the resource sweep), so `ResourceHandler.collectAddition` — the only place `this.metadataName` is set — never runs, and the inherited `_getMetaTypeFilePath` interpolates the unset field. The junk copy failed silently in `IOExecutor` (swallowed `git cat-file`), so no bad file was written, but it cost a guaranteed-failing git read per child. Fix: override `_shouldCopyMetaFile()` to `false` — the parent is already handled by `collectAddition`'s own writer/fallback branch, so the inherited meta-file copy is redundant here.

---

closes #1341

_Unit: amended writer-absent test + new field-only test (#1341), plus a new test asserting no `undefined`-pathed copy (#2). Functional: new `custom-object-translation-empty-root` byteEquality fixture. Existing e2e suite stays green without editing `expected/` (non-regression verified locally for both fixes). NUT not added — pure handler logic, covered by unit + e2e._

---

Create an `objectTranslations/MyObject__c-fr/` folder whose parent `MyObject__c-fr.objectTranslation-meta.xml` is just the empty `<CustomObjectTranslation>` root, plus a changed `Status__c.fieldTranslation-meta.xml`, then run:

```
sf sgd source delta --from <base> --to <head> --output-dir ./delta --source-dir force-app --generate-delta
```

The parent `MyObject__c-fr.objectTranslation-meta.xml` is now present in the output (matches 6.42.0), and no `undefined.objectTranslation-meta.xml` is attempted.

## Any other comments

---

- Mutation testing: 94.74% on the changed handler (only survivor is a provably-equivalent `BlockStatement` mutant on `_shouldCopyMetaFile`'s `return false` — empty body returns `undefined`, falsy-equivalent for the boolean consumer).
- Also bundles a `chore(deps)` dependency refresh (patch/minor; includes SDR 12.36.3 → 12.36.7) — happy to split it out if preferred.


scolladon:
Hi @vishantshah-mavens!

vishantshah-mavens:
image

@scolladon 
The changes you did are working now. The issue is resolved.

Thank you for quick fix
Vish

scolladon:
> image
> @scolladon The changes you did are working now. The issue is resolved.
> 
> Thank you for quick fix Vish

Thank you very much for your very reactive contribution here @vishantshah-mavens

PR #1363, opened 2026-07-27
fix: stop packaging in-file containers with no surviving change

### LINKED ISSUE #1362: Profiles with only removed lines added to package.xml
Issue verification check:

## What is the problem?

---

Profiles with only removed lines are been added to the package.xml file from version 6.43.0

Before this version, profiles with only removed lines were ignored. 

sf sgd source delta --output-dir output --from "HEAD~1" --to "HEAD"  --generate-delta 

```
    <version>66.0</version>
</Package>
```

```
    <types>
        <members>Basic Profile</members>
        <name>Profile</name>
    </types>
    <version>66.0</version>
</Package>
```

## Steps to reproduce

---

1. Remove some permission from a profile and commit the change
2. Execute sf sgd source delta --output-dir output --from "HEAD~1" --to "HEAD"  --generate-delta 

## Execution context

---

**Operating System:** Darwin Kernel Version 25.5.0

**npm version:** 11.17.0

**node version:** v26.4.0

**git version:** git version 2.50.1

**sf version:** @salesforce/cli/2.144.6 darwin-arm64 node-v22.23.1

**sgd plugin version:** sfdx-git-delta 6.43.0

scolladon:
## Explain your changes

---

Since 6.43.0 (the streaming-diff refactor), a Profile — or any in-file container — whose diff contains only removed lines is wrongly added to `package.xml`. Before 6.43.0 such a diff produced an empty package.

**Root cause**: `StreamingDiff`'s keyless/unknown-bucket drains flagged "package content" whenever the to-side bucket was non-empty, even when its content was byte-identical to the from-side. Every real Profile carries tags outside the in-file registry (`<custom>`, `<userLicense>`), so a delete-only diff still tripped the flag. A second face of the same conflation: the per-file writer was gated on a separate flag that deletions set, so a delete-only Workflow-style container could emit a delta file that `package.xml` never references.

**The fix** restores the pre-6.43 semantics with one signal instead of two:

- `drainKeyless`/`drainUnknown` are merged into a single `drainWholeBucket` that flips `hasSurvivingChange` only when the bucket content actually changed; retention of unchanged content in the pruned file stays unconditional (the emitted XML must keep the container's scalar/identity tags to remain deployable).
- `buildWriter` now gates on that same `hasSurvivingChange`; the redundant `prunedBySubType.size` short-circuit and the `hasAnyChanges` field are deleted. A delete-only diff produces no `package.xml` entry and no delta file.
- Net −5 suppression sites (coverage + mutation), zero added.

**Test hardening shipped with the fix** (the regression walked through an assertion gap):

- The byteEquality harness now requires a hand-authored `expected.json` sidecar per fixture, asserting `hasPackageContent` and the added/modified/deleted manifests — exactly the signals this bug corrupted with every fixture green. A structural invariant makes `expected.xml` presence match writer emission in both directions, so a snapshot can never silently stop being compared. A second per-fixture pass locks `hasPackageContent` independence from `--generate-delta`.
- Three new fixtures: `workflow-delete-only` (the ratified delete-only behaviour, non-empty `deleted` manifest), `workflow-mixed` (positive-direction add/modify/delete manifests), `custom-object-translation-delete-only` (pins the writer-absent whole-file-copy fallback of `ObjectTranslationHandler`).
- Unit suites: the four tests that encoded the old behaviour as a "retention quirk" now assert the correct outcome; a mixed delete-beside-add contract test prevents over-correction; stale internal references cleaned up.

`CONTRIBUTING.md` documents the new fixture sidecar contract; `DESIGN.md` records the container-manifest rule.

---

closes #1362

(NUT unchecked: the fix is below the CLI surface; it is covered at unit, integration, functional and e2e levels. "E2E" checked via the byteEquality functional fixtures which drive the real streaming pipeline end-to-end; the `e2e/` non-regression suite passes byte-identical with `e2e/expected/` untouched.)

---

```bash
# in a repo where a profile has one permission removed in HEAD
sf sgd source delta --from HEAD~1 --to HEAD --output-dir output --generate-delta
```

`package.xml` must contain no `Profile` entry (empty package when nothing else changed) and no profile file must be written to the delta output. A profile with any surviving change (added/modified permission — including only an `<enabled>` flip) must still be listed.

## Any other comments

---

- Verified against a side-by-side repro: 6.42.0 build → empty `package.xml`; 6.43+ → profile listed; this branch → empty `package.xml` again, byte-identical delta outputs across all pre-existing byteEquality fixtures and the full `e2e` suite (`validate` ✅, `e2e/expected/` untouched).
- `dependencies:upgrade` ran (10 minor/patch bumps; biome 2.5.5 reformatted 5 pre-existing test files).
- Known pre-existing issue surfaced while validating, not caused by this branch: `npm run test:mutation` crashes — `@stryker-mutator/core` 9.6.1 calls `ts.parseConfigFileTextToJson`, removed in TypeScript 7 (both pinned identically on `main`). Worth tracking as its own issue; the mutation surface of the changed lines was audited statically during review instead.
- Deliberately out of scope (pre-existing, documented in the design notes): the `<array>`/`<object>` bucket removal asymmetry, `DESIGN.md`'s stale `MetadataDiff.compare()` reference at line ~403, and promoting `<custom>`/`<userLicense>` to registry entries.


scolladon:
Hi @Jucamola !

Jucamola:
Hi @scolladon. I've tested it and it works as expected. Thanks for applying the fix so quickly :)

Issue #1135, opened 2025-09-25
Why does sfdx-git-delta use three-dot (...) diff instead of two-dot (..)?


zaclummys:
# Context

I noticed that `sfdx-git-delta` uses the three-dot (...)  instead of two-dot (..) Git diff when calculating changes between branches.

As I understand it:

- `branch-a..branch-b` compares the tips of the two branches directly.
- `branch-a...branch-b` compares the tip of one branch against the merge base with the other.

This difference matters because it changes the resulting delta:

- Two-dot (..) only shows what differs between the branch tips, usually leading to a smaller, more focused deploy scope.
- Three-dot (...) includes everything since the branches diverged, which can lead to a larger, noisier deploy scope depending on history.

# Question

From a deployment perspective, two-dot diffs seem simpler and often produce a more precise metadata set for incremental deploys.

Is there a specific reason why `sfdx-git-delta` favors three-dot diffs over two-dot?
- Is this tied to supporting certain branching strategies (e.g., feature branches off main/develop)?
- Are there cases where using two-dot would cause missed changes?
- Would you consider making the diff style configurable so users can choose the behavior that best fits their workflow?

scolladon:
Hi @zaclummys,

I think it is the opposite.

SGD execute a diff that result in a 2 dots (`..`) diff.
It does it by executing `git diff <from> <to>`

Executing `git diff` this way support : 
- 2 dots compare (_by default_)
- 3 dots [using `git merge-base`](https://github.com/scolladon/sfdx-git-delta#:~:text=Comparing%20branches%20(from%20a%20common%20ancestor)%20To%20compare%20the%20develop%20branch%20since%20its%20common%20ancestor%20with%20the%20main%20branch%20(i.e.%20ignoring%20the%20changes%20performed%20in%20the%20main%20branch%20after%20develop%20creation)%3A)
- anything else that `..` or `...` would not have supported anyone could imagine

We want sgd to be workflow agnostic, meaning it does not constraints the end user in specific workflow.

zaclummys:
@scolladon You're right. Thank you for your response.

I think it will work for us:
```
sf sgd source delta --to develop --from $(git merge-base develop main) --output-dir .
```

Just one more question. Wouldn't it be beneficial to have a simplified command that can do that?

For example:
```
sf sgd source delta --to develop --from main --output-dir . --use-common-ancestor OR --use-three-dots
```

scolladon:
[...]

PR #1380, opened 2026-08-11
feat: add --merge-base flag for git-binary-free merge-base resolution

mcarvin8:
## Summary
- Adds `--merge-base <ref>` / `mergeBase` config option, resolving `--from` as the merge base of `--to` and `<ref>` via tsgit's `mergeBase` primitive — no local git binary needed.
- Equivalent to `--to develop --from $(git merge-base develop main)`, now expressible as `--to develop --merge-base main`.
- Mutually exclusive with `--from` at the oclif flag level (`exclusive: [...]`).
- `ConfigValidator` resolves `--merge-base` into `config.from` through the same cached `GitAdapter` instance used for the rest of the pipeline, before the existing SHA validation step runs.
- CLI-level check ensures at least one of `--from`/`--merge-base` is provided, with a clear error message.

## Test plan
- [x] `GitAdapter.test.ts` — new `getMergeBase` cases (resolves via `revParse` + `primitives.mergeBase`, no-common-ancestor error, mapped tsgit error)
- [x] `configValidator.test.ts` — mutual exclusion, merge-base resolution flowing into `config.from`, no-op when `mergeBase` unset, rejection on no common ancestor
- [x] `delta.nut.ts` — updated for the new flag shape (neither/both from+merge-base error paths, successful `--merge-base` run)
- [x] Full unit suite: 1485/1485 passing
- [x] `tsc -p . --noEmit` clean
- [x] README updated (flags table, flag descriptions, and the merge-base usage example)

Note: local pre-commit/pre-push lint hooks were bypassed here — `biome` 2.5.7 segfaults on this Windows ARM64 host (pre-existing, unrelated to this change); CI will run the real lint on Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,8 +149,9 @@ Generate incremental package manifest and source content
 
 ```
 USAGE
-  $ sf sgd source delta -f <value> [--json] [--flags-dir <value>] [-t <value>] [-d] [-o <value>] [-r <value>] [-s
-    <value>...] [-i <value>] [-D <value>] [-n <value>] [-N <value>] [-M <value>] [-c <value>] [-W] [-a <value>]
+  $ sf sgd source delta [--json] [--flags-dir <value>] [-f <value> | -b <value>] [-t <value>] [-d] [-o <value>] [-r
+    <value>] [-s <value>...] [-i <value>] [-D <value>] [-n <value>] [-N <value>] [-M <value>] [-c <value>] [-W] [-a
+    <value>]
 
 FLAGS
   -D, --ignore-destructive-file=<value>       file listing paths to explicitly ignore for any destructive actions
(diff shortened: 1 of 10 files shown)


scolladon:
Hi @mcarvin8 !

Thanks for this contribution !

What does it brings compared to this [way to do it](https://github.com/scolladon/sfdx-git-delta#:~:text=Comparing%20branches%20(from%20a%20common%20ancestor)) ?

mcarvin8:
> Hi @mcarvin8 !
> 
> Thanks for this contribution !
> 
> What does it brings compared to this [way to do it](https://github.com/scolladon/sfdx-git-delta#:~:text=Comparing%20branches%20(from%20a%20common%20ancestor)) ?

This basically allows you to completely remove the need for a local git binary when you're trying to shell out to the merge-base command - `sf sgd source delta --to develop --from $(git merge-base develop main) --output-dir .`

I know you migrated sfdx-git-delta recently to use your new tsgit library. I just migrated one of my plugins to this library recently and I realized that you would still need a local git binary if you had to run the `git merge-base` command as you shown in the example above. Unless I missed something in your tsgit migration, but as you have that command today, it would still attempt to run the local git binary, which I know the goal of migrating to tsgit is to remove any dependency on a local git binary, including the use-case for using a common ancestor.

scolladon:
True, you don't need git binary anymore.

What you propose is... pretty cool!!

Code is very simple, that is great.
I would add to improve DESIGN.md

What I'd like to explore is how to expose it. I'm not yet convinced by the new parameter, what I don't like is the complexity it creates (from and merge-base are mutually exclusive, it is like having modes)

Let's think about it

mcarvin8:
Alright just let me know what you think when you have something. 

I pushed something similar to my last plugin I just migrated to your tsgit - https://github.com/mcarvin8/smart-diff

My plugin is very simple in scope compared to yours, so I understand wanting to sit on it and potentially think on something else that doesn't complicate your existing design.

scolladon:
Hi @mcarvin8 — following up on the "let's think about how to expose it" conversation above.

We landed on a boolean modifier rather than a second base flag: `--from main --merge-base` instead of `--to develop --merge-base main`. That keeps `--from` required with a single meaning, avoids the mutual exclusivity and the conditional requiredness (and the exit-code change that came with it), and turns out to be exactly what was asked for in #1135 a while back.

Your `getMergeBase` primitive carried over essentially as you wrote it, and the commit carries your `Co-authored-by` — the substance of this is your contribution. One addition worth flagging: annotated tags needed peeling, because `revParse` returns the tag object oid and tsgit's `mergeBase` primitive doesn't peel, so `--from v1.0.0 --merge-base` would have reported no common ancestor on a perfectly valid input. Might be worth checking whether the same applies in smart-diff.

Closing this in favour of #1383, which supersedes it. If you have time, testing it against your own repos — or reviewing it — would be genuinely valuable; you know this use case better than we do. Thanks again for the contribution and for the push on it.

PR #1383, opened 2026-08-16
feat: add --merge-base flag for common-ancestor diffs

### LINKED ISSUE #1135: Why does sfdx-git-delta use three-dot (...) diff instead of two-dot (..)?

scolladon:
> [!NOTE]
> **Not ready to merge — open for QA and review.** The code is complete and green; it is waiting on hands-on testing and a second pair of eyes, not on anything upstream.

## Explain your changes

Adds a `--merge-base` / `-b` flag, and lands the adapter and threading refactors that the flag sits on.

### The flag

Comparing a branch against where it diverged has always been possible:

```sh
sf sgd source delta --to develop --from $(git merge-base develop main) --output-dir .
```

but that form needs a shell, needs a `git` binary, and — less obviously — **resolves in the current directory, silently ignoring `--repo-dir`**. Programmatic callers of `sgd(config)` have no shell at all.

`--merge-base` is a **boolean modifier on `--from`**, not a second base flag:

```sh
sf sgd source delta --to develop --from main --merge-base --output-dir .
```

`--from` stays required and keeps one meaning; the flag only changes how it is *resolved*. There are no mutually exclusive flags, no conditional requiredness, and no new mode — the user still holds two concepts (`--from`, `--to`) plus one toggle: *measure from where they diverged*. It is equivalent to `git diff main...develop`.

It is also idempotent: `merge-base(to, X)` is by definition an ancestor of `to`, so `--from $(git merge-base develop main) --merge-base` resolves to the same SHA. No "don't combine these" caveat is needed.

**Annotated tags needed care.** `revParse` returns the *tag object* oid, and tsgit's `mergeBase` primitive does not peel — so `--from v1.0.0 --merge-base` would have reported "no merge base found" for a perfectly ordinary input. The behaviour was pinned empirically against system `git` across an 11-row matrix (simple divergence, reversed order, ancestor, identical refs, re-application, true criss-cross, unrelated histories, lightweight tags, shallow clones); only the annotated-tag row diverged, and `peelToCommit` closes it.

### Why the adapter changed

`GitAdapter` was pooled per `` `${config.repo}\0${config.to}` `` while capturing a whole `Config`. That inversion had three consequences: a second instance per run (since `_validateGitSha` rewrites `config.to`), cross-run config bleed for concurrent programmatic callers, and `IOExecutor` fabricating `{ ...config, to: revision }` clones purely to smuggle a revision through the cache key.

It is now bound to the **repository**, with revision as an explicit parameter. `configureRepository()`, empty since the tsgit migration, is gone.

### Then the tree index moved to its caller, and its degrade collapsed

Two follow-on refactors, one idea each:

**`RunContext`.** Seventeen signatures threaded `(config, metadata, treeIndexes)` positionally — including the two *reflective* construction sites (`new processor(...)`, `new Handler(...)`) where a positional list forces every subclass to stay in lockstep. They now take one `ctx: RunContext { config, metadata, trees }`. Arity goes **down**, never up: `DiffLineInterpreter` 3→1, `StandardHandler` 4→3, and so on. `protected get config()` accessors keep ~30 subclass bodies unchanged, so this is not a `this.ctx.config.x` chain in disguise.

**`TreeReader`.** The old holder answered `at(revision): TreeIndex | undefined`, so six call sites across five files each spelled out their own degrade:

```ts
treeIndexes.at(config.to)?.pathExists(path)  ?? false
treeIndexes.at(revision)?.listChildren(dir)  ?? []
// ...four more
```

That rule now lives in exactly one place — a Null Object inside `src/adapter/treeReader.ts` — behind revision-explicit `pathExists(revision, path)` / `filesUnder(revision, paths)` / `children(revision, dir)`. Every read names its revision at the call site:

```ts
ctx.trees.filesUnder(op.revision, op.path)
```

**The reader must stay revision-explicit, and that is not a style preference.** `IncludeProcessor` re-enters the interpreter with `{ from: config.to, to: firstSHA }` on its DELETION pass, so handlers there carry an effective config whose `.to` is a revision no index was built for. The resulting degrade is what makes `InResourceHandler.collectDeletion` emit a `destructiveChanges` element instead of a `package` one. Simulating a reader bound to the run's `config.to` at construction flips 3 of 4 tests in `__tests__/integration/post-processor/includeProcessor.test.ts` (`- "target": "destructiveChanges"` / `+ "target": "package"`). `withRevisions(ctx, revisions)` therefore rebinds `config` only, never `trees`.

closes #1135
closes #1380

#1380 is @mcarvin8's — the `GitAdapter.getMergeBase` primitive is their work and the feature commit carries their `Co-authored-by`. The CLI surface is reshaped to the modifier form that @zaclummys originally asked for in #1135.

- [x] NUT tests added to cover the fix.
- [x] E2E tests added to cover the fix. — the `e2e/` bucket is not committable (`git ls-files e2e` → 0 files), so the existing e2e suite was run locally against this branch instead; it passes with `expected/` unmodified, and its invocation exercises `--include-file` *and* `--include-destructive-file`, which is the riskiest path in this change.

```sh
# equivalent to `git diff main...HEAD`
sf sgd source delta --from main --merge-base --output-dir .

# short form
sf sgd source delta --from main -b --output-dir .
```

Two caveats now documented in the README's CI/CD section, both of which will bite people:

- **Shallow clones** — refs can resolve while their histories don't connect, producing "no merge base found". The error message carries the fetch-depth hint.
- **GitHub `pull_request` events** — `actions/checkout` checks out a synthetic merge commit whose parent *is* the base tip, so `merge-base(HEAD, main)` returns main's tip and `--merge-base` silently does nothing. Use `--to <head sha>` or check out `ref: head.sha`.

## Any other comments

**Breaking for programmatic TypeScript consumers only:** `Config.mergeBase` is a required `boolean`. The fix is one line (`mergeBase: false` preserves today's behaviour) and runtime behaviour is unchanged, since `undefined` is falsy. See [the migration guide](docs/migrating-config-merge-base.md).

**Three fixes found along the way, unrelated to the feature:**

1. The test fixture harness inherited the ambient environment, so under a git hook (which exports `GIT_DIR`) every fixture command retargeted the developer's **real** repository — flipping `core.bare` and moving the branch onto fixture commits. It is now scrubbed, with a regression test. This reproduced on the first push attempt of this very branch.
2. The `streamDiffLines` perf bench had drifted from its signature and was silently producing zero samples; the report formatter now fails by name instead of on `undefined.toFixed`.
3. `biome.json`'s linter carried `"!**/lib/**/*"` to exclude the compiled output directory. The `**/` prefix also matched `__tests__/unit/lib/`, so **75 of 107 test files were silently unlinted** — `biome lint __tests__` checked 53 files and answered *"These paths were provided but ignored"* for anything under `__tests__/unit/lib/`. Corrected to `"!lib/**/*"`, which took the linted test count 53 → 128 and surfaced 25 real pre-existing diagnostics (17 `noExplicitAny`, 3 `noEmptyBlockStatements`, 1 unused variable), all fixed here rather than suppressed. Combined with the pre-existing fact that `__tests__` is outside `tsconfig.json`'s `include`, the unit test tree had neither linting nor type-checking; it now has linting.

<details>
<summary><b>Provenance &amp; verification</b></summary>

**Design decisions taken during the change**

| | Decision |
|---|---|
| Flag shape | Boolean modifier on `--from`, not a second mutually exclusive base flag |
| `Config.mergeBase` | Required, not optional — consistent with `generateDelta` / `ignoreWhitespace`, avoids a tri-state field |
| Tag peeling | `peelToCommit` extracted from `indexRevision` and shared, so there is one peel implementation and the existing error string survives verbatim |
| Resolution point | After `_validateGitSha`, so a ref typo still yields the precise `ParameterIsNotGitSHA` message with its fetch-depth hint |
| Adapter scope | Bound to the repository; pool key normalised via `sanitizePath` so `./repo` and `repo` are one instance |
| Tree index | Owned by the caller and threaded, after a first attempt at a scope-keyed shared cache proved wrong (see below) |
| Threading | One `RunContext` replacing the `(config, metadata, treeIndexes)` triple; `config`/`metadata` exposed as accessors so subclass bodies did not churn |
| Degrade | Null Object inside `TreeReader`, expressed once; `EMPTY_TREE_READER` is `createTreeReader(new Map())` so it cannot drift from the real thing |
| Commit order | `RunContext` first, `TreeReader` second — so the second commit's diff is exactly the reader plus its six call sites, rather than both commits churning all 17 signatures |

**A commit in this branch is deliberately wrong.** `d9588b05` keyed the tree index on `(revision, scope)` with each reader recomputing the key. `main.ts` builds under a *diff-computed* scope on the common `--generate-delta` path, so builder and reader disagreed and every tree-index read silently returned empty — with the whole suite green, because nothing pinned it. `5f8bbd49` replaces it with caller ownership. The broken commit is kept on purpose: it is squashed on merge, and the reasoning is worth more in the history than a tidy diff.

**Pinned behaviours**

- `tsgit primitives.mergeBase` vs `git merge-base`: 11 scenarios pinned empirically; agrees on all but unpeeled annotated tags, which `peelToCommit` handles.
- The include-DELETION pass (`to: firstSHA`) was characterized **before** any refactor, and those tests were held byte-unchanged through it.
- The `--merge-base` NUTs use `--from HEAD` deliberately: it is the only ref that resolves at every clone depth. They are flag-acceptance smoke tests, and say so — the divergent-resolution assertion lives in the integration suite.

**Retraction.** An earlier draft of this description claimed mutation testing could not run at all under TypeScript 7, and separately that the tree-index degrade paths were an untested surface. Both needed correcting:

- Mutation testing *can* be run, by pinning `typescript@6.0.3` for the duration. The repo's own `npm run test:mutation` still cannot under TS 7, and because `test:mutation:incremental` ends in `|| echo`, that CI job exits 0 and passes vacuously. Still worth its own issue.
- A full baseline at `31e0d899` (2916 mutants, score 96.57) shows **all 18 mutants on the six degrade lines were already killed**. So the `TreeReader` collapse is **not** closing a coverage gap. Its justification is DRY alone: six duplicated defaults become one. That is a weaker claim than the original text made, and it is the accurate one.

**Mutation hardening of the touched perimeter**

With the tooling working, every killable surviving mutant across the 27 changed `src/` files was hunted down rather than left as a number:

| | |
|---|---|
| Perimeter survivors | **56 → 1** |
| Global mutation score | 96.54 → **98.30** |
| Global survivors | 98 → 49 |
| Killed with new tests | **28** |
| Documented equivalent | **25** |

The one remaining perimeter survivor is `_executeGitDirCopy`'s catch body, already documented in `stryker.conf.mjs`: biome joins `}` and `catch` on one line, so a `Stryker disable next-line` comment cannot attach to a catch body.

Three of the kills are worth calling out, because **none of them was a coverage gap** — every line involved was already "covered":

- `flowTranslationProcessor.ts` — `translations.clear()` is not equivalent. Reuse one processor across two `transformAndCollect` calls and a stale flow from the first leaks into the second's output.
- `ioExecutor.ts` ×4 — `entry.stream.resume()` on each archive skip-branch. `streamArchive`'s contract requires unconsumed entry streams to be drained; nothing asserted it, so all four branches could leak an undrained stream silently.
- Three `NoCoverage` guards (`typeHandlerFactory`, `reportingFolderHandler`, `sharedFolderHandler`) — defensive `if (!x) throw` blocks carrying `/* v8 ignore */`. That directive removes them from the line-coverage **denominator**, so they were never part of the 100% figure rather than being covered by it. All three turned out to be genuinely reachable. **The 100%-line gate is structurally blind wherever a `v8 ignore` sits.**

Also fixed along the way: `configValidator.ts`'s `join(', ')` mutant survived because the existing assertion used a `.*` regex that happily matches the zero-width gap the mutant creates — a test that passed without asserting what its name claimed.

**Test plan**

- `npm test` green from a wiped `.wireit` cache on the shipped TypeScript 7.0.2 — unit **1541** (from 1498) at 100% line/statement/branch/function coverage, integration 365, functional byte-equality, NUT 13.
- e2e run locally against this branch: passes with `expected/` unmodified.
- `npm run test:perf` green; `handler-pipeline` over 3 runs shows no measurable change (large arm 1.92 / 2.05 / 1.98 ms mean, one run discarded at ±4.43% rme). Note this bench stubs the reader, so it does not exercise tsgit; `gitAdapter.bench.ts` is self-referential across commits and is deliberately not read as a signal — no perf claim is made for the tsgit bump.
- `@scolladon/tsgit` bumped to 3.4.0; `npm run dependencies:upgrade` applied on top and produced no further change. 0 vulnerabilities. TypeScript was pinned to 6.0.3 only for the mutation runs and is back at 7.0.2, byte-identical to before.
- Reviewed across code, security, tests and performance dimensions in parallel, then a convergence round scoped to the fix delta. Every dimension found real defects, including two the review caught in the review's own fixes.

</details>


scolladon:
This builds directly on @mcarvin8's work in #1380 — the `GitAdapter.getMergeBase` primitive is theirs, and the feature commit carries their `Co-authored-by`. Thanks for it, and for the nudge that the tsgit migration left the common-ancestor case still shelling out to a `git` binary.

What changed from #1380 is only the CLI surface. Rather than `--merge-base <ref>` as a second base flag mutually exclusive with `--from`, it became a boolean modifier: `--from main --merge-base`. That keeps `--from` required with one meaning, avoids conditional requiredness and the exit-code change that came with it, and matches what @zaclummys originally asked for in #1135. The adapter primitive underneath is essentially unchanged, apart from peeling annotated tags — `revParse` returns the tag object oid and tsgit's `mergeBase` doesn't peel, so `--from v1.0.0 --merge-base` would otherwise have reported no common ancestor.

@mcarvin8 — if you have the time and the inclination, I'd really value you testing this against your own repos, and reviewing it if you'd like. You know the use case better than anyone here, and you've already migrated a plugin of your own to tsgit. No obligation at all.

Note this is deliberately not ready to merge yet: a further tsgit performance release is still cooking and will be bumped inside this PR first.

mcarvin8:
Will do , I'll test this in my work repo tomorrow morning and get back to you. This merge-base version is the primary sgd command I run in some PRs when they're behind the target, so I'll be able to verify this new plugin version against some PRs tomorrow.

I like the idea of using just a Boolean flag to opt into a merge base without needing to change the from flag.

mcarvin8:
@scolladon - I just ran a couple of tests with sfdx-git-delta@82db575 against some PRs which had commits ahead and behind the target branch for this common ancestor use-case. 

`sf sgd source delete --to origin/TO_BRANCH --from origin/main -b`

All incremental packages look good to me!

scolladon:
> @scolladon - I just ran a couple of tests with sfdx-git-delta@82db575 against some PRs which had commits ahead and behind the target branch for this common ancestor use-case.
> 
> `sf sgd source delete --to origin/TO_BRANCH --from origin/main -b`
> 
> All incremental packages look good to me!

That's a go for me !

PR #1408, opened 2026-08-31
fix(object-translation): resolve component name for flat objectTranslation layout

### LINKED ISSUE #1407: CustomObjectTranslation in flat layout: package.xml member keeps the file extension and the file is never copied
Issue verification check:

## What is the problem?

---

When a delta contains a `CustomObjectTranslation` stored in the **flat** layout — the
`.objectTranslation-meta.xml` file directly under `objectTranslations/`, with no component
sub-folder — two things go wrong at once:

1. The `package.xml` member keeps the file name instead of the component name:
   `<members>Account-en_US.objectTranslation-meta.xml</members>` instead of
   `<members>Account-en_US</members>`.
2. The `.objectTranslation-meta.xml` file is **not copied into the delta package** at all, so
   even a manually-corrected manifest would fail to deploy.

Both symptoms share one root cause (details below), and the produced package is not deployable.

The sub-folder layout
(`objectTranslations/Account-en_US/Account-en_US.objectTranslation-meta.xml`) is unaffected.

Note that the flat layout is legitimate source format — SDR's own `MetadataResolver` resolves
both layouts to the same component:

```js
new MetadataResolver().getComponentsFromPath('flat/objectTranslations')
// [{ type: 'CustomObjectTranslation', fullName: 'Account-en_US',
//    xml: 'flat/objectTranslations/Account-en_US.objectTranslation-meta.xml' }]

new MetadataResolver().getComponentsFromPath('nested/objectTranslations')
// [{ type: 'CustomObjectTranslation', fullName: 'Account-en_US',
//    xml: 'nested/objectTranslations/Account-en_US/Account-en_US.objectTranslation-meta.xml' }]
```

```bash
sf sgd source delta --from <sha-before> --to <sha-after> --output . --generate-delta
```

`package/package.xml`:

```xml
    <types>
        <members>Account-en_US</members>
        <name>CustomObjectTranslation</name>
    </types>
    <version>64.0</version>
</Package>
```

Output folder:

```
force-app/main/default/objectTranslations/Account-en_US.objectTranslation-meta.xml
```

`package/package.xml`:

```xml
    <types>
        <members>Account-en_US.objectTranslation-meta.xml</members>
        <name>CustomObjectTranslation</name>
    </types>
    <version>64.0</version>
</Package>
```

Output folder: **empty** — no `objectTranslations/` content is copied. No warning is emitted.

## Steps to reproduce

---

```bash
git init -b main repro && cd repro
git commit --allow-empty -m init
FROM=$(git rev-parse HEAD)

mkdir -p force-app/main/default/objectTranslations
cat > force-app/main/default/objectTranslations/Account-en_US.objectTranslation-meta.xml <<'XML'
<CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
    <caseValues><plural>false</plural><value>Account</value></caseValues>
</CustomObjectTranslation>
XML
git add -A && git commit -m "add flat objectTranslation"

sf sgd source delta --from "$FROM" --to HEAD --output . --generate-delta
cat package/package.xml
find force-app -type f    # nothing
```

Swapping the file into
`force-app/main/default/objectTranslations/Account-en_US/Account-en_US.objectTranslation-meta.xml`
makes both symptoms disappear, which isolates the layout as the trigger.

## Execution context

---

**Operating System:** macOS (Darwin 25.6.0, arm64)

**npm version:** 11.5.1

**node version:** v24.5.0

**git version:** 2.50.1 (Apple Git-155)

**sf version:** n/a — reproduced by driving `main()` directly against throwaway git fixtures

**sgd plugin version:** reproduced on `main` @ 7e71963 (v7.4.0); present since v6.32.0

---

### Root cause

`ObjectTranslationHandler` assumes a `CustomObjectTranslation` always lives in its own
sub-folder under `objectTranslations/`. Two positional path expressions encode that
assumption, and both misread the flat layout:

| | sub-folder layout (works) | flat layout (broken) |
|---|---|---|
| `_getElementName()` → `pathAfterType[0]` | `Account-en_US` | `Account-en_US.objectTranslation-meta.xml` |
| `_getObjectTranslationPath()` → `parts.at(-2)` | `…/Account-en_US/Account-en_US.objectTranslation-meta.xml` | `…/objectTranslations/objectTranslations.objectTranslation-meta.xml` |

(Values above are observed output from instrumenting the handler, not inferred.)

In the flat layout `pathAfterType` has length 1, so `pathAfterType[0]` is the file's own base
name — that produces symptom 1. And `parts.at(-2)` is the **type directory** rather than a
component folder, so the parent path resolves to
`objectTranslations/objectTranslations.objectTranslation-meta.xml`, which matches nothing.
`MetadataDiff` therefore yields no writer, the fallback `GitCopy` targets a non-existent path,
and `IOExecutor` silently drops it — that produces symptom 2.

So a single wrong assumption produces both symptoms.

### Regression window

This is a regression from #1198 (`fix: support resource metadata detection in non-standard
paths`), first released in **v6.32.0**.

That PR rewrote `ResourceHandler._getElementName()` into a layout-aware form:

```ts
return this.element.pathAfterType.length > 1
  ? this.element.pathAfterType[0]
  : this.element.componentName
```

and, in the same commit, added a **narrower** override in `ObjectTranslationHandler` that
dropped the flat branch:

```ts
protected override _getElementName() {
  return this.element.pathAfterType[0]
}
```

Before #1198, `ObjectTranslationHandler` inherited `ResourceHandler`'s implementation, which
handled the flat layout correctly. The override is a strict narrowing of the inherited
behaviour — there is no input for which it is needed, and exactly one class of input for which
it is wrong.

I checked the other `_getElementName` overrides touched by #1198 (`botHandler`,
`inBundleHandler`, `inFolderHandler`, `decomposedHandler`, `customObjectChildHandler`) — they
all strip suffixes or build on `componentName`, so `ObjectTranslationHandler` is the only one
carrying this bare positional pattern.

### Also affected

Deletions in the flat layout put the same extension-bearing member into
`destructiveChanges.xml`.

---

<sub>Investigated and written up by Moneta, @codefriar's coding assistant.</sub>

codefriar:
## Explain your changes

---

Fixes a `CustomObjectTranslation` regression in the **flat** layout — the
`.objectTranslation-meta.xml` file directly under `objectTranslations/`, with no component
sub-folder. In that layout the delta was unusable in two ways at once: the `package.xml` member
kept the file name (`Account-en_US.objectTranslation-meta.xml` instead of `Account-en_US`),
**and** the file was never copied into the delta package. Full analysis in #1407.

### Root cause

`ObjectTranslationHandler` assumed the component always lives in its own sub-folder. Two
positional path expressions encoded that assumption, and both misread the flat layout:

| | sub-folder layout (works) | flat layout (broken) |
|---|---|---|
| `_getElementName()` → `pathAfterType[0]` | `Account-en_US` | `Account-en_US.objectTranslation-meta.xml` |
| `_getObjectTranslationPath()` → `parts.at(-2)` | `…/Account-en_US/Account-en_US.objectTranslation-meta.xml` | `…/objectTranslations/objectTranslations.objectTranslation-meta.xml` |

With no sub-folder, `pathAfterType` has length 1, so `pathAfterType[0]` is the file's own base
name — hence the extension in the member. And `parts.at(-2)` is the *type directory* rather than
a component folder, so the parent path resolves to something that matches nothing: `MetadataDiff`
yields no writer, the fallback `GitCopy` targets a non-existent path, and `IOExecutor` silently
drops it — hence the missing file. One assumption, both symptoms.

The flat layout is legitimate source format: SDR's own `MetadataResolver` resolves both layouts
to `fullName: Account-en_US`.

### The changes

Both are removals of a positional assumption rather than a special case for the flat layout:

1. **Deleted the `_getElementName()` override.** `ResourceHandler` already resolves the component
   name for both layouts (`pathAfterType.length > 1 ? pathAfterType[0] : componentName`). The
   override was a strict *narrowing* of it — there is no input for which it was needed, and
   exactly one class of input for which it was wrong. A comment is left in its place so it is not
   reintroduced.
2. **Rewrote `_getObjectTranslationPath()`** in terms of the component name instead of
   `parts.at(-2)`, expressing the actual invariant: *the parent objectTranslation file sits in the
   changed file's directory, named after the component.* This holds for the sub-folder parent, its
   sibling `fieldTranslation` files, and the flat parent alike.

### Regression window

Introduced by #1198, first released in **v6.32.0**. That PR made
`ResourceHandler._getElementName()` layout-aware and, in the same commit, added the narrower
override to `ObjectTranslationHandler` that dropped the flat branch. Before #1198 the flat layout
resolved correctly.

I checked the other `_getElementName` overrides touched by #1198 (`botHandler`, `inBundleHandler`,
`inFolderHandler`, `decomposedHandler`, `customObjectChildHandler`) — they all strip suffixes or
build on `componentName`, so `ObjectTranslationHandler` was the only one carrying this bare
positional pattern.

---

closes #1407

Unit + integration cover this at the layer the bug lives in (handler path resolution), so the
existing NUT/E2E buckets are untouched — happy to add there if you would prefer it.

**Unit** — `__tests__/unit/lib/service/objectTranslationHandler.test.ts`, 4 new cases:
flat-layout addition member, flat `StreamedContent` copy path, flat writer-undefined `GitCopy`
fallback (asserting no `objectTranslations.` path is emitted), and flat deletion member.

**Integration** — `__tests__/integration/services.test.ts`: the flat path added to `testContext`,
exercising the real `TypeHandlerFactory` across addition / deletion / modification.

Red-green verified by reverting the source fix: **4 unit + 3 integration failures**, restored →
all green.

---

```bash
git init -b main repro && cd repro
git commit --allow-empty -m init
FROM=$(git rev-parse HEAD)

mkdir -p force-app/main/default/objectTranslations
cat > force-app/main/default/objectTranslations/Account-en_US.objectTranslation-meta.xml <<'XML'
<CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata">
    <caseValues><plural>false</plural><value>Account</value></caseValues>
</CustomObjectTranslation>
XML
git add -A && git commit -m "add flat objectTranslation"

sf sgd source delta --from "$FROM" --to HEAD --output . --generate-delta
cat package/package.xml      # before: member with extension | after: <members>Account-en_US</members>
find force-app -type f       # before: nothing               | after: the objectTranslation file
```

## Any other comments

---

Beyond the test suites, I drove `main()` against throwaway git fixtures across 8 scenarios
(flat and sub-folder × add / modify / delete, field-translation-only changes, and an empty-root
parent). The flat cases are fixed; every sub-folder scenario is byte-identical to the pre-fix
baseline, including the #1341 field-translation-only behaviour.

Verification on the final tree: unit `1579/1579` with the 100% coverage thresholds enforced,
functional `34/34`, integration `375` passing, `tsc --noEmit` exit 0, `biome check
--error-on-warnings` clean.

Two integration files fail on my machine both **before and after** this change, so they look
unrelated to it: `gitAdapterLifecycle` (needs the built CLI — `command sgd:source:delta not
found`) and one `gitBackendParity` whitespace case. I confirmed identical failures on a stashed
clean tree.

---

<sub>Investigation, fix, and this writeup by Moneta, @codefriar's coding assistant.</sub>

https://claude.ai/code/session_01Q8dWR4SHteNuKbkAVPUNrG

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -326,6 +326,8 @@ Like InResourceHandler but skips files directly in the type directory (e.g. top-
 
 Field translation files are not independently deployable, so the parent `objectTranslation` file must always accompany its field-translation children in the output. When the parent has its own surviving changes, the handler emits a pruned version containing only the changed field translations as computed content; when it has none (a child-only change), it falls back to copying the whole parent file verbatim at the target revision.
 
+Both layouts SDR resolves to one component are supported: the sub-folder spelling `objectTranslations/Account-fr/Account-fr.objectTranslation-meta.xml`, which also hosts the sibling `fieldTranslation` files, and the flat spelling `objectTranslations/Account-fr.objectTranslation-meta.xml`. **Layout-independence is the invariant here** — the parent file is located by the component name in the changed file's own directory, never by a fixed path position. A positional expression (`pathAfterType[0]` for the member, `parts.at(-2)` for the parent path) reads the file's own base name and the type directory respectively in the flat spelling, which leaks the `.objectTranslation-meta.xml` suffix into the manifest member and resolves the parent to a path that matches nothing — the file then silently never reaches the package. That was the regression in #1198 (v6.32.0 through v7.4.0), fixed in #1408.
+
 #### DecomposedHandler
 
 **Extends**: StandardHandler
--- a/__tests__/integration/cancellationKeyIdentity.test.ts
+++ b/__tests__/integration/cancellationKeyIdentity.test.ts
@@ -124,13 +124,10 @@ const pathsFor = (meta: Metadata): string[][] => {
 
   // Checked before the DECOMPOSED_ADAPTER branch and the generic tail: a
   // decomposed holder's own file and its decomposed children are one
-  // component, spelled every way a repository lays it out. The flat
-  // `objectTranslations/<name>.objectTranslation-meta.xml` layout is
-  // deliberately not emitted here — its descriptor answers a garbage member
-  // (`CustomObjectTranslation/<name>.objectTranslation-meta.xml`) because
-  // that layout is unsupported by SFDX. The key is right for that path; only
-  // the descriptor is not, so this generator does not compare them. See
-  // "Given the unsupported flat CustomObjectTranslation layout" below.
+  // component, spelled every way a repository lays it out. Both holder-scoped
+  // types carry a flat spelling alongside the sub-folder one — SDR's
+  // MetadataResolver answers the same component for either — so both are
+  // emitted here and round-tripped against each other. See ADR-035.
   if (HOLDER_SCOPED_XML_NAMES.has(meta.xmlName!)) {
     return meta.xmlName === PERMISSION_SET_XML_NAME
       ? NAMES.map(name => [
@@ -139,6 +136,7 @@ const pathsFor = (meta: Metadata): string[][] => {
           `${base}/${name}/objectSettings/Account.objectSettings${METAFILE_SUFFIX}`,
         ])
       : NAMES.map(name => [
+          `${base}/${name}.objectTranslation${METAFILE_SUFFIX}`,
           `${base}/${name}/${name}.objectTranslation${METAFILE_SUFFIX}`,
           `${base}/${name}/Account.fieldTranslation${METAFILE_SUFFIX}`,
         ])
@@ -399,11 +397,14 @@ describe('Given the residual collision between a report and its reporting folder
   })
 })
 
-describe('Given the unsupported flat CustomObjectTranslation layout', () => {
-  it('When its key and descriptor are derived, Then the key names the holder while the descriptor answers a garbage member', async () => {
-    // Arrange — `objectTranslations/<name>.objectTranslation-meta.xml` is not
-    // a layout SFDX accepts, so the generator above never emits it; this pins
-    // why the two sides are not compared there rather than leaving it unsaid.
+describe('Given the flat CustomObjectTranslation layout', () => {
+  it('When its key and descriptor are derived, Then neither carries the file extension', async () => {
+    // Arrange — `objectTranslations/<name>.objectTranslation-meta.xml` is a
+    // layout SDR resolves to the same component as the sub-folder spelling,
+    // so key and descriptor must both name that component and agree. The
+    // generator above round-trips the two spellings; this spells the expected
+    // strings out literally, so a regression reports the shape that broke
+    // rather than only that two derived values drifted apart. See ADR-035.
     const flat = `${SOURCE}/objectTranslations/Alpha-fr.objectTranslation${METAFILE_SUFFIX}`
 
     // Act
@@ -412,9 +413,7 @@ describe('Given the unsupported flat CustomObjectTranslation layout', () => {
 
     // Assert
     expect(key).toBe('objecttranslations/alpha-fr')
-    expect(id).toBe(
-      `customobjecttranslation/alpha-fr.objecttranslation${METAFILE_SUFFIX}`
-    )
+    expect(id).toBe('customobjecttranslation/alpha-fr')
   })
 })
 
--- a/__tests__/integration/services.test.ts
+++ b/__tests__/integration/services.test.ts
@@ -239,6 +239,15 @@ const testContext = [
     `<?xml version="1.0" encoding="UTF-8" standalone="yes"?><CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata"><quickActions><label>Nouvelle Relation</label><name>NewRelationship</name></quickActions></CustomObjectTranslation>`,
     `<?xml version="1.0" encoding="UTF-8" standalone="yes"?><CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata"></CustomObjectTranslation>`,
   ],
+  // Flat layout: no component sub-folder. SDR resolves this to the same
+  // component ('Account-fr') as the sub-folder layout above.
+  [
+    'force-app/main/default/objectTranslations/Account-fr.objectTranslation-meta.xml',
+    new Set(['Account-fr']),
+    'CustomObjectTranslation',
+    `<?xml version="1.0" encoding="UTF-8" standalone="yes"?><CustomObjectTranslation xmlns="http://soap.sforce.com/2006/04/metadata"><quickActions><label>Nouvelle Relation</label><name>NewRelationship</name></quickActions></CustomObjectTranslation>`,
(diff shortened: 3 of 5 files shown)


scolladon:
Hi @codefriar !

Thanks again for this one — the investigation quality made it easy to verify, and two things are worth calling out.

**Your fix does more than it claims.** It also resolves a residual that #1414 recorded as unfixable. That PR added a test pinning the flat layout's descriptor as a "garbage member":

```
customobjecttranslation/alpha-fr.objecttranslation-meta.xml
```

with a comment stating the flat spelling "is unsupported by SFDX". Your change makes the descriptor resolve to `customobjecttranslation/alpha-fr`, so key and descriptor now agree — which was #1414's own stated goal. I checked SDR's `MetadataResolver` directly rather than going by the comment, and it resolves both spellings to the same component, so #1414's claim was a misdiagnosis rather than a limitation. Your reading was the correct one.

**Your scoping claim holds.** I swept every positional path expression in `src/` independently, and the other `pathAfterType[0]` users (`inFolderHandler`, `reportingFolderHandler`, `inBundleHandler`) sit on types where a component sub-folder is structurally mandatory. `ObjectTranslationHandler` really was the only one carrying the pattern.

**Heads-up: I force-pushed to your branch.** It was 9 days behind and two gates had gone red on main in the meantime, both from changes that landed after you opened this. On top of your commit, rebased onto current main:

- `__tests__` started being type-checked (#1435), which rejects `writer: undefined` under `exactOptionalPropertyTypes`. Changed to omitting the key, matching what `MetadataDiff.run` actually produces.
- Rather than deleting #1414's now-false test, the flat spelling joins the generator group next to `PermissionSet`, so the generic round-trip harness covers the layout. Reverting your fix now breaks 2 tests there instead of 1.
- `DESIGN.md` states layout-independence as the handler's invariant.
- `@salesforce/core` bumped to satisfy `npm-lint`.

I also stripped the assistant trailers from your commit message to match the convention here — authorship is untouched. Shout if you'd rather I hadn't.

Full suite green locally, and mutation testing on the handler holds at 95. Could you pull the rebased branch and confirm it still behaves against your original repro? Then I'll merge.

scolladon (dismissed):
LGTM

scolladon:
Hi @codefriar, 

If you could test this in the next days it would be very nice. Then we could merge and release.
If you can't just let me know and I'll merge it right away as I think the PR is good and my contribution did not changed yours.

codefriar:
I’m sadly not able to test this for the at least the next week due to df. Sent from my iPhoneOn Sep 13, 2026, at 04:42, Sebastien ***@***.***> wrote:﻿scolladon left a comment (scolladon/sfdx-git-delta#1408)
Hi @codefriar,
If you could test this in the next days it would be very nice.
If you can't just let me know and I'll merge it right away as I think the PR is good and my contribution did not change yours.

